PR #56, opened 2026-04-23
feat: :sparkles: create website

martonvago:
# Description

This PR updates from the template (and updates dependencies like Sprout) and builds the website.

The one issue is the width of the tables, but that's not specific to this repo. Raised https://github.com/seedcase-project/seedcase-flower/issues/301 about it.

Closes #42

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`


resources/adverse_events.qmd
martonvago, line 1: Is it okay to have the `resources` folder in the root?

_quarto.yml
```diff
@@ -0,0 +1,41 @@
+project:
+  type: website
+
+book:
+  title: "ON LiMiT Feasibility Study Data"
+  # TODO: Set the URL of the website
```
martonvago, line 6: Do we have this already?

resources/adverse_events.qmd
lwjohnst86: Hmm, let's put everything into `docs/`. The landing page `index.qmd` in root can than point to the metadata docs, along with other information (e.g. more about ON-LiMiT, partner institutions, etc).

lwjohnst86: I also wonder if maybe these should be generated as part of the CI/CD workflow and not saved here in the repo. What do you think?

.flower.toml
```diff
@@ -0,0 +1,2 @@
+style = "quarto-resource-tables"
+output-dir = "."
```
lwjohnst86, line 2:
```suggestion
output-dir = "docs"
```

_quarto.yml
lwjohnst86: We do not.

lwjohnst86 (changes requested):
Some high level comments

martonvago:
Oops forgot about this PR 😅

.flower.toml
martonvago: With this change, `index.qmd` would also be generated in `docs`. And then Quarto builds everything at `.../_site/docs/index.html`. 

Maybe if we generate the docs as part of the CI/CD workflow, then we can generated them with `output-dir = "."` and it won't matter.

Or we could move `index.qmd` to the root after Flower runs.

lwjohnst86: Hm, Quarto would build at the root, not `docs/`. We would have to add an `index.qmd` to the root.

martonvago: Ah, so you're thinking that we should add _another_ `index.qmd` at the root?

resources/adverse_events.qmd
martonvago: I think that would be neat! I'm just thinking that it might not play nicely with the existing `reusable-build-docs-with-python` workflow because I don't know if that will see the files generated by Flower (without sharing them).

.flower.toml
martonvago: Okay, so if the files will be generated in the GitHub workflow, then they can be generated in the root. Done in https://github.com/seedcase-project/template-data-package/pull/219

lwjohnst86: Hm, not sure I understand here. I think it would help to have a landing page that *isn't* built from Flower. So include a "Welcome" index page that has some basic details about the feasibility data that gives more context to the metadata documentation.

martonvago: Ah, okay! That wasn't clear to me, sorry 😅

martonvago:
Closing in favour of #70

Issue #32, opened 2026-03-11
Extract choices for categorical as number from data dictionary


lwjohnst86:
E.g. gender is 1 and 2, so needs to be converted

martonvago:
In the metadata, checkbox fields are one field with a series of options.
In the data, each option gets its own column (with a value of checked or not checked), as in the symptom example above.

Which representation do we want to use: do we want to 
~~1. transform the data to match the metadata
➡️ `"symptom": ["cough", "fever"]`
Slight problems with this: the options can be full sentences; the DP schema has no way of declaring that all elements in an array must come from a given set of options~~
2. transform the metadata to match the data
➡️ expand each checkbox REDCap field into multiple `FieldProperties`, one for each option

PR #66, opened 2026-05-05
feat: :sparkles: download raw data

### LINKED ISSUE #63: Download data as raw from REDCap
Pull from REDCap direct into `raw/redcap/<instrument-name>/<timestamp>.csv.gz` without any modifications. Store as gzipped or some other compression to save space.

martonvago:
# Description

This PR is a first-pass at downloading raw data.

Some open questions:
- Should we save raw data as CSV or JSON (see issue)?
- Given the discussion about wanting to provide the metadata (also the data?) in multiple languages, should we leave checkbox and radio fields as numbers (see #32)? 
- Should we fetch the data instrument by instrument or fetch everything at once and split by instrument locally?
- Should we structure the code thematically (how it's done here) or by layer (separate file for REDCap client and separate files for other logic)?
- Should the raw data be tracked by GitLFS or gitignored? If it goes in the same folder as the tracked data, it would make sense for it to be tracked and not ignored.

Closes #63 

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`


scripts/redcap_data.py
martonvago, line 1: I split code thematically into data code and metadata code, rather than pulling out a `REDCapClient` class. But we can do that if we think it's nicer, especially if we think it will be reused in other data packages.

scripts/redcap_data.py
```diff
@@ -0,0 +1,75 @@
...
+from pathlib import Path
+
+import requests
+import seedcase_soil as so
+from dotenv import load_dotenv
+
+load_dotenv()
+
+
+def get_instruments() -> list[str]:
```
martonvago: So that later I can iterate through all the instruments.

scripts/redcap_data.py
```diff
@@ -0,0 +1,75 @@
...
+        raise RuntimeError("REDC_CPH_API_KEY environment variable is not set.")
+
+    data = {
+        "token": token,
+        "content": "record",
+        "action": "export",
+        "format": "json",
+        "type": "flat",
+        "csvDelimiter": "",
+        "forms[0]": instrument,
```
martonvago: We can ask for the data by instrument. This saves us the trouble of splitting it ourselves, but it also means that we make 80ish API requests, which can be a little slow, especially if there's lots of data. It doesn't look like REDCap has any limits in place for serving this many requests, but I haven't looked into this beyond testing it.

scripts/redcap_data.py
```diff
@@ -0,0 +1,75 @@
...
+    """Gets the data for a specific instrument from REDCap as JSON."""
+    token = os.environ.get("REDC_CPH_API_KEY")
+    if not token:
+        raise RuntimeError("REDC_CPH_API_KEY environment variable is not set.")
+
+    data = {
+        "token": token,
+        "content": "record",
+        "action": "export",
+        "format": "json",
```
martonvago: We can put `csv` here, but then we get all the data as a long string, which feels brittle to me. Especially because the `csvDelimiter` field doesn't seem to be respected all the time.

scripts/redcap_data.py
```diff
@@ -0,0 +1,75 @@
...
+        "exportDataAccessGroups": "false",
+        "returnFormat": "json",
+    }
+    response = requests.post("https://redcap.regionh.dk/api/", data=data, timeout=60)
+    response.raise_for_status()
+    return response.json()
+
+
+def save_instrument_data(instrument: str):
+    """Saves instrument data to `raw/redcap/<instrument>/<timestamp>.json.gz`."""
```
martonvago: Saving as JSON here for the reasons mentioned in the issue.

scripts/redcap_data.py
```diff
@@ -0,0 +1,75 @@
...
+    timestamp = datetime.now().strftime("%Y-%m-%dT%H-%M-%S")
+    file_path = Path("raw") / "redcap" / instrument / f"{timestamp}.json.gz"
+    file_path.parent.mkdir(parents=True, exist_ok=True)
+    with gzip.open(file_path, "wt", encoding="utf-8") as f:
+        json.dump(data, f, indent=2, ensure_ascii=False)
+
+
+if __name__ == "__main__":
+    instruments = get_instruments()
+    so.pretty_print(f"Found {len(instruments)} instruments.")
```
martonvago: These messages are pretty useful to keep track of what's going on.

scripts/redcap_data.py
```diff
@@ -0,0 +1,75 @@
...
+        "format": "json",
+        "returnFormat": "json",
+    }
+    response = requests.post("https://redcap.regionh.dk/api/", data=data, timeout=30)
+    response.raise_for_status()
+    instruments = response.json()
+    return so.fmap(instruments, itemgetter("instrument_name"))
+
+
+def get_instrument_data_from_redcap(instrument: str) -> list[dict[str, str]]:
```
lwjohnst86: The ending `from_redcap` I don't think is necessary, since it's in the `redcap_data.py` file.

scripts/redcap_data.py
lwjohnst86: Related to #63 try with CSV first.

scripts/redcap_data.py
lwjohnst86: Hmmmmm, so we could make a single request and get all the data? It might be a lot faster and less resource intensive to pull the full data down and split locally. Especially if the instrument name is in the CSV download.

Actually, yes, let's do that, since we'll end up doing a lot of processing of the data to get the instruments to match the resources (e.g. `vas` as one resource rather than the multiple instruments).

scripts/redcap_data.py
```diff
@@ -0,0 +1,75 @@
...
+    response = requests.post("https://redcap.regionh.dk/api/", data=data, timeout=60)
+    response.raise_for_status()
+    return response.json()
+
+
+def save_instrument_data(instrument: str):
+    """Saves instrument data to `raw/redcap/<instrument>/<timestamp>.json.gz`."""
+    data = get_instrument_data_from_redcap(instrument)
+    timestamp = datetime.now().strftime("%Y-%m-%dT%H-%M-%S")
+    file_path = Path("raw") / "redcap" / instrument / f"{timestamp}.json.gz"
```
lwjohnst86: When pull the entire data, this `instrument` path could be removed.

scripts/redcap_data.py
```diff
@@ -0,0 +1,75 @@
...
+        json.dump(data, f, indent=2, ensure_ascii=False)
+
+
+if __name__ == "__main__":
+    instruments = get_instruments()
+    so.pretty_print(f"Found {len(instruments)} instruments.")
+
+    for instrument in instruments:
+        save_instrument_data(instrument)
+        so.pretty_print(f"Saved data for instrument: {instrument!r}.")
```
lwjohnst86: If we make a single request and download, this isn't needed.

lwjohnst86 (changes requested):

scripts/redcap_data.py
martonvago: Okay! But then I shouldn't split them by "raw" instrument at this point, right? Because they will be split to match the resources later as you say.

scripts/redcap_data.py
```diff
@@ -0,0 +1,51 @@
...
+    if not token:
+        raise RuntimeError("REDC_CPH_API_KEY environment variable is not set.")
+
+    data = {
+        "token": token,
+        "content": "record",
+        "action": "export",
+        "format": "csv",
+        "type": "flat",
+        "csvDelimiter": ";",
```
martonvago, line 26: This option doesn't seem to be respected, but the `regionh` server uses `;`, so I put it there.

.gitignore
```diff
@@ -27,7 +27,6 @@ __pycache__/
 *.py[cod]
 
 # Raw data files and folders
-*.csv
```
martonvago, line 30: Track CSV files using Git LFS

scripts/redcap_data.py
lwjohnst86: Hmmm, that's weird 🤔 but good to be explicit what the delimiter is

lwjohnst86 (approved):
Nice 👍

PR #70, opened 2026-05-18
feat: :sparkles: add website

martonvago:
# Description

This PR adds the website.

Closes #42 

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`


lwjohnst86 (changes requested):
There are more changes here than just adding the website. Could you split the changes up to be atomic?

martonvago:
@lwjohnst86 it's updating from the template -- can those changes not stay together? Or should I do one straight update and one PR for customising the update?

lwjohnst86:
Yea, even when updating from the template, it's good to keep things small and atomic. That way it's also easy to check if anything happened during the update that needs to be fixed. That can be difficult to check/review when a PR has 32 files changed.

lwjohnst86 (approved):

PR #86, opened 2026-05-20
chore: :wrench: remove toc to avoid table overlap

### LINKED ISSUE #82: [bug] Resource table description and toc overlap
See https://onlimit-study.github.io/feasibility-data/docs/#resources---scroll down and see how the table of content ("On this page") overlaps the table content bc the table content overflows the page.

Screenshot: 

Image

signekb:
# Description

This doesn't fix the issue of tables having different widths bc Quarto doesn't break/wrap the long variable names, but it does avoid the tables overlapping with the toc. And the toc doesn't really add anything too meaningful on these pages anyway, e.g., on the package overview page, all the headers in the toc can be seen when viewing the page:

image

Closes #82 

This PR needs a quick review.

### DIFF
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -27,6 +27,7 @@ website:
 
 format:
   onlimit-theme-html:
+    toc: false
     theme:
       - brand
     # TODO: Uncomment if using the goatcounter website visitor counter


lwjohnst86 (approved):
That's true, we don't need TOCs in general for this website.

Because this is a data package, `fix`, `feat`, and `refactor` apply only to changes to the metadata and data, not the config or docs. I changed it to `chore`

signekb:
Ah, yes, ofc, I thought of it as a docs repo. Thanks for changing it 👍

PR #87, opened 2026-05-22
style: :lipstick: set Data Package section grid and change sidebar style to "docked"

signekb:
# Description

Updates "after" photo below with no custom css solution. 

This changes the layout like so: 

**Before**: 

image

**With these changes**:

image

## Checklist

- [ ] Ran `just run-all` (fails bc I `main.py` can't run)


docs/custom.css
```diff
@@ -0,0 +1,36 @@
+body.floating.fullcontent .page-columns {
+  grid-template-columns:
+    /*
+    Quarto does not have options to remove outer margins, so we have to redefine the entire grid.
```
joelostblom: It's not enough to set the left and right hand side widths to 0px, as per [this doc page](https://quarto.org/docs/output-formats/page-layout.html#html-page-layout)?

signekb:
This, however, could be something we would like to address directly in Flower instead, see seedcase-project/seedcase-flower#301

joelostblom (changes requested):
I like the change in appearance from these changes. Only downside I can see is if the left sidebar is perceived to "jump" position between separate pages, but I think the benefits are greater than that issues.

I wonder if it would be advantageous to have this (or a similar) css as the default for the web pages with tables on them?

docs/custom.css
signekb: I tried that, setting `margin-width: 0px`, but it doesn't really do much:

```
format:
  onlimit-theme-html:
    grid:
      margin-width: 0px
```

image

With a combination of making the sidebar and body wider, we can get something a bit more like the desired result: 

```
format:
  onlimit-theme-html:
    grid:
      sidebar-width: 400px
      body-width: 900px 
```

image

As I understand it, these setting don't touch the white space per say, but only expands the components, thereby reducing the white space.

Only caveat is that this solution sets the width globally, i.e., also for the landing page:

image

signekb:
Thanks for the review, Joel 🌷 

Related to the sidebar "jump": The sidebar is the same across all pages in the Data Package section, but yes, the content body is stays the default width on all other pages (such as the landing page). That doesn't really bother me. Changing the body width across all pages isn't a problem to me either, but **only** as long as this page only has a landing page and the data package section; if more pages will be added later, the wider body might hurt readability.

For comparison, with these changes, the landing page and data package section looks like this:

image

image

docs/custom.css
lwjohnst86: The grid only impacts all of the website if it is used in the `_quarto.yml` doc. But if you put the settings in `docs/_metadata.yml`, they only impact pages under `docs/`

e.g. in `docs/_metadata.yml`

```
format:
  onlimit-theme-html:
    toc: false
    grid:
      sidebar-width: 550px
      body-width: 1000px
```

Along with using in `_quarto.yml`

```
...
  sidebar:
    - id: package
      style: "docked"
      contents:
        - text: "Package"
          href: docs/index.qmd
        - auto: docs/resources/
```

can produce something similar without impacting the other pages and without needing custom css.

joelostblom:
Great! Yes, the content width doesn't really bother me either, but if it is easily possible to change it to only apply to the docs as per Luke's comment in the other thread, then that would be preferred. If it is not easily possible, I'm also good with this PR in its current form.

docs/custom.css
signekb: Ah, nice! Thanks, Luke! 👌

signekb:
Thanks, @lwjohnst86 ! 👌 💅  I've removed the custom css again. With this, the resource tables still have different widths though, but that's minor. 

  Edit: One thing that still bothers me is the column width (which I tried to fix with custom css). With long variable names, the name column can be quite wide bc Quarto's default is to not break words. This can result in the Description col being quite narrow, e.g.: 

image

lwjohnst86:
@signekb  we know that all tables in `docs/resources` have 4 columns, you can use `tbl-colwidth` in a `docs/resources/_metadata.yml` file to set the exact widths

signekb:
@lwjohnst86 Hmm, that doesn't seem to do anything, unfortunately 🤔 

One way it could be fixed is to add a little bit of custom css (could be directly in `_metadata.yml`), like so: 

```
tbl-colwidths: [25, 25, 10, 40] <-- line added 
format:
  onlimit-theme-html:
    toc: false
    grid:
      sidebar-width: 550px
      body-width: 1000px
    include-in-header: <-- section added
      text: |
        <style>main table { table-layout: fixed; } main table td, main table th { overflow-wrap: break-word; }</style>
```

lwjohnst86:
Oh, you want to break words. I misunderstood. No, for now, we don't need that. The `name` column shouldn't break and the `title` column will eventually be actual words separated by ` ` not `_` so this issue won't really be a problem then.

lwjohnst86 (approved):
This is good for now 👍 Once we've tidied up the metadata we can look more at the styling.

PR #89, opened 2026-05-22
refactor: :recycle:  convert column time suffixes to its own variable in VAS

### LINKED ISSUE #45: Need to convert VAS resources into one resource
There's multiple VAS resources (90 min, 60 min, etc). This should be tidied so that VAS is in one data resource, not many. E.g. the variable names shouldn't have a unit in that, the unit should be a column of its own.

joelostblom:
# Description

@martonvago Sorry this took longer than expected. Here is an attempt make changes to the script and the resulting differences in the `datapackage.json` file. All `_XXmin` suffixes are now removed and instead included as a variable. Don't hesitate to let me know if I missunderstood something from the issue.

Closes #45

This PR needs a quick/an in-depth review.

## Checklist

- [x] Ran `just run-all`


scripts/redcap_dict_to_properties.py
```diff
@@ -42,6 +47,64 @@ def dictionary_to_properties(
...
+    fields, seen_vas_fields = result
+    field_name = field["field_name"]
+
+    if field["form_name"] != "vas":
+        return fields + [field], seen_vas_fields
+
+    if field_name in seen_vas_fields:
+        return result
+
+    return fields + [field], seen_vas_fields | {field_name}
```
```suggestion
    return (fields + [field], seen_vas_fields | {field_name})
```

Just to visually show its a tuple.

What's the `|` doing here?

scripts/redcap_dict_to_properties.py
```diff
@@ -65,6 +128,23 @@ def _form_to_resource(
...
     )
+    default_fields = [event_field, center_field]
+    primary_key = ["event"]
+
+    if form_name == "vas":
+        time_field = sp.FieldProperties(
+            name="time",
+            title="Time",
+            type="integer",
+            description="The time in minutes after the meal when the VAS was recorded.",
```
```suggestion
            description="The time in minutes before or after the meal when the specific VAS measurement was recorded.",
```

scripts/redcap_dict_to_properties.py
```diff
@@ -65,6 +128,23 @@ def _form_to_resource(
             enum=["Copenhagen", "Aarhus", "Odense"],
         ),
     )
+    default_fields = [event_field, center_field]
+    primary_key = ["event"]
+
+    if form_name == "vas":
+        time_field = sp.FieldProperties(
+            name="time",
```
```suggestion
            name="minutes_from_meal",
```

scripts/redcap_dict_to_properties.py
```diff
@@ -65,6 +128,23 @@ def _form_to_resource(
...
+            title="Time",
```
```suggestion
            title="Minutes from the meal",
```

scripts/redcap_dict_to_properties.py
```diff
@@ -65,6 +128,23 @@ def _form_to_resource(
...
+            type="integer",
+            description="The time in minutes after the meal when the VAS was recorded.",
+            categories=VAS_TIMEPOINTS,
+            constraints=sp.ConstraintsProperties(
+                required=True,
+                enum=VAS_TIMEPOINTS,
+            ),
+        )
+        default_fields.append(time_field)
+        primary_key.append("time")
```
```suggestion
        primary_key.append("minutes_from_meal")
```

datapackage.json
```diff
@@ -27069,10 +27069,10 @@
...
     },
     {
-      "name": "vas_120_min",
-      "path": "resources/vas_120_min/data.parquet",
-      "title": "vas_120_min",
-      "description": "vas_120_min",
+      "name": "vas",
+      "path": "resources/vas/data.parquet",
+      "title": "vas",
+      "description": "vas",
```
lwjohnst86: I couldn't figure out where these were added. These should be different from the `name`.

scripts/redcap_dict_to_properties.py
```diff
@@ -42,6 +47,64 @@ def dictionary_to_properties(
...
+    )
+
+
+def _normalise_vas_time_resource_field(field: dict[str, str]) -> dict[str, str]:
+    if not _is_vas_time_resource_field(field):
+        return field
+
+    return {
+        **field,
+        "field_name": VAS_TIME_FIELD_PATTERN.sub("", field["field_name"]),
```
lwjohnst86: You can also remove the prefix `vas_` from the fields.

lwjohnst86 (changes requested):
Nicely reduces the number of tables we have 🎉 just some comments and changes.

scripts/redcap_dict_to_properties.py
```diff
@@ -9,6 +10,9 @@
 
 In = TypeVar("In")
 Out = TypeVar("Out")
+VAS_TIMEPOINTS = [30, 60, 90, 120, 180, 240]
```
martonvago: What about `vas_minus10`, should that also be included here?

scripts/redcap_dict_to_properties.py
joelostblom: It's the union operators for sets; I switched to the method name instead for clarity since we don't often perform set operations.

scripts/redcap_dict_to_properties.py
joelostblom: Well spotted! I only looked for the `*min` suffix and missed this one; added!

joelostblom (commented):
Thanks for the comments; updated accordingly!

scripts/redcap_dict_to_properties.py
```diff
@@ -94,19 +181,37 @@ def _form_to_resource(
...
+
+
+def _get_resource_description(form_name: str) -> str:
+    if form_name == "vas":
+        return (
+            "Visual analogue scale measurements recorded at multiple timepoints "
+            "relative to the meal."
+        )
+
+    return form_name
```
joelostblom, line 219: These two functions give more specific names for the title and description field in `datapackage.json`

martonvago (dismissed):
🎉 👏👏👏

scripts/redcap_dict_to_properties.py
```diff
@@ -94,19 +181,37 @@ def _form_to_resource(
...
-            fields=[event_field, center_field] + form_fields + checkbox_fields,
+            primary_key=primary_key,
+            fields=default_fields + form_fields + checkbox_fields,
         ),
     )
 
 
+def _get_resource_title(form_name: str) -> str:
+    if form_name == "vas":
+        return "Visual analogue scale measurements"
```
lwjohnst86, line 207: @K-Beicher is this an accurate title for this resource?

datapackage.json
```diff
@@ -27104,884 +27104,126 @@
...
             ]
           },
           {
-            "name": "vas_hunger_fasted_minus10min",
-            "title": "vas_hunger_fasted_minus10min",
+            "name": "hunger",
+            "title": "hunger",
             "type": "number",
-            "description": "Visual analogue scale. How hungry do you feel, at time minus 10 min Question: Hvor sulten er du?. Slider scale labels: Jeg er slet ikke sulten |  | Jeg har aldrig været mere sulten",
+            "description": "Visual analogue scale. How hungry do you feel Question: Hvor sulten er du?. Slider scale labels: Jeg er slet ikke sulten |  | Jeg har aldrig været mere sulten",
```
lwjohnst86: The `"Visual analogue scale. "` can be removed from these descriptions. Also, can you add a regex replacer to put a period before `Question:` if there is no period?

lwjohnst86 (changes requested):
Looking very nice, only two suggested changes (plus asking @K-Beicher about the title).

datapackage.json
joelostblom: Done in cf61caf

lwjohnst86 (approved):

PR #90, opened 2026-05-27
refactor: ♻️ join SEFNC weeks resources

### LINKED ISSUE #46: `sefnc` has many resources, but it should only have one
The units are in the table name and the variable names (e.g. _v10 or _v6 and _week12). The columns shouldn't have units in their names

joelostblom:
# Description

This joins the SEFNC visit-specific resources into a single `sefnc` resource instead of having multiple resources with `_vN` suffixes. 

I opted to repeat some of the `vas_` functions here, because it didn't seem like much was gained in generalizing already. See https://github.com/onlimit-study/feasibility-data/issues/91 for details.

Closes #46

## Checklist

- [x] Ran `just run-all`


scripts/redcap_dict_to_properties.py
```diff
@@ -13,6 +13,13 @@
 VAS_TIMEPOINTS = [-10, 30, 60, 90, 120, 180, 240]
 VAS_TIME_FORM_PATTERN = re.compile(r"^vas_(minus10|(30|60|90|120|180|240)_?min)$")
 VAS_TIME_FIELD_PATTERN = re.compile(r"(_fasted)?_(minus10|30|60|90|120|180|240)min$")
+SEFNC_WEEKS = [0, 12, 52]
+SEFNC_FORM_WEEKS = {
+    "sefnc_baseline_v4": 0,
+    "sefnc_week12_v6": 12,
+    "selfefficacy_for_nutrition_change_sefnc_week_52": 52,
+}
```
joelostblom, line 21: I went with the week values `0` (baseline), `12`, and `52` and removed the suffixes `_v6` and `_v10`. Initially I thought that the suffixes would be the weeks, but the don't line up with the description, so maybe they stand for something like "visit" instead (but maybe it should have been a "b" in Danish...)? Maybe @K-Beicher knows?

scripts/redcap_dict_to_properties.py
```diff
@@ -109,6 +116,79 @@ def _remove_vas_time_from_annotation(annotation: str) -> str:
...
+        "field_annotation": _remove_sefnc_week_from_annotation(
+            field["field_annotation"]
+        ),
+    }
+
+
+def _normalise_sefnc_field_name(field_name: str) -> str:
+    return SEFNC_WEEK_FIELD_PATTERN.sub("", field_name).replace(
+        "sefnc_ubusy_schedule", "sefnc_busy_schedule"
+    )
```
joelostblom, line 159: Some of the fields used the name `sefnc_ubusy_schedule` and some left out the `u` and called it `sefnc_busy_schedule`. I don't know if the `u` is a valuable part to keep or just a type that is fine to remove as I did here.

datapackage.json
```diff
@@ -23991,7 +24010,7 @@
             "name": "sefnc_when_busy",
             "title": "sefnc_when_busy",
             "type": "string",
-            "description": "Participant's confidence in following their diet when they have a heavy workload. Self-reported by participant. Repeated at baseline (V4), week 12 (V6) and week 52 (V10).",
```
joelostblom, line 23994: This line mentions that the `v` suffix does not represent the week.

scripts/redcap_dict_to_properties.py
martonvago: I also assumed "v" means "visit". Especially because in the data the `event` for `_v6` fields is `besg_6_arm_1`, for `_v10` `besg_10_arm_1`, etc. (Does this also mean that the week info is already implicit in the `event` field?)

scripts/redcap_dict_to_properties.py
```diff
@@ -38,7 +45,7 @@ def dictionary_to_properties(
     redcap_fields: list[dict[str, str]],
 ) -> list[sp.ResourceProperties]:
     """Converts REDCap data dictionary to Data Package resources."""
-    redcap_fields = _join_vas_time_resources(redcap_fields)
+    redcap_fields = _join_sefnc_week_resources(_join_vas_time_resources(redcap_fields))
```
martonvago: Maybe you're already thinking about this in #91, but maybe it would be nicer to have these steps sequential instead of nested. That would allow us to add or remove these transformations in a more modular way.

scripts/redcap_dict_to_properties.py
```diff
@@ -109,6 +116,79 @@ def _remove_vas_time_from_annotation(annotation: str) -> str:
...
+        **field,
+        "field_name": _normalise_sefnc_field_name(field["field_name"]),
+        "form_name": "sefnc",
+        "field_annotation": _remove_sefnc_week_from_annotation(
+            field["field_annotation"]
+        ),
+    }
+
+
+def _normalise_sefnc_field_name(field_name: str) -> str:
```
martonvago, line 156: Should we also drop the `sefnc_` prefix like with `vas_`?

scripts/redcap_dict_to_properties.py
lwjohnst86: Exactly. These are the times I wish Python had a pipe 😢 


```suggestion
    redcap_fields = _join_sefnc_week_resources(redcap_fields)
    redcap_fields = _join_vas_time_resources(redcap_fields)
```

scripts/redcap_dict_to_properties.py
lwjohnst86: Yes! ☺️

scripts/redcap_dict_to_properties.py
lwjohnst86: @K-Beicher?

scripts/redcap_dict_to_properties.py
```diff
@@ -109,6 +116,79 @@ def _remove_vas_time_from_annotation(annotation: str) -> str:
     ).strip()
 
 
+def _join_sefnc_week_resources(
```
lwjohnst86, line 133: We may have to consider a better way of organizing the scripts, this will start getting big...

scripts/redcap_dict_to_properties.py
```diff
@@ -109,6 +116,79 @@ def _remove_vas_time_from_annotation(annotation: str) -> str:
...
+def _deduplicate_sefnc_fields(fields: list[dict[str, str]]) -> list[dict[str, str]]:
+    deduplicated_fields, _ = reduce(
+        _append_if_new_sefnc_field,
+        fields,
+        ([], set()),
+    )
+    return deduplicated_fields
+
+
+def _append_if_new_sefnc_field(
```
lwjohnst86, line 176: This could probably be generalised in some way no?

scripts/redcap_dict_to_properties.py
```diff
@@ -153,6 +233,21 @@ def _form_to_resource(
         default_fields.append(time_field)
         primary_key.append("minutes_from_meal")
 
+    if form_name == "sefnc":
+        week_field = sp.FieldProperties(
+            name="week",
+            title="Week",
```
lwjohnst86: Would this be week from the first visit?

datapackage.json
```diff
@@ -23812,11 +23812,30 @@
...
+              12,
+              52
+            ]
+          },
           {
             "name": "sefnc_when_tired",
             "title": "sefnc_when_tired",
             "type": "string",
-            "description": "Participant's confidence in following their diet when feeling tired. Self-reported by participant. Baseline (V4).",
+            "description": "Participant's confidence in following their diet when feeling tired. Self-reported by participant.",
```
lwjohnst86: "Self-reported by participant" Could probably be removed, as that information is better placed in the resource description.

lwjohnst86 (changes requested):
Nice, just some comments/changes :)

scripts/redcap_dict_to_properties.py
joelostblom: Yes I opened https://github.com/onlimit-study/feasibility-data/issues/91 for discussing that.

scripts/redcap_dict_to_properties.py
joelostblom: I opened #91 for discussing generalization since I thought it was bigger than just this PR and I wanted to keep the PR small.

scripts/redcap_dict_to_properties.py
joelostblom: Yup, I agree it's nicer not to nest.

A pipe would be nice too. If we really want something like that, we could consider taking on one of the functional python libraries as a dependency maybe https://toolz.readthedocs.io/en/latest/api.html#toolz.functoolz.pipe (would maybe give more the other functional capabilities as well, but also seems a bit unnecessary if we are likely planning to switch to Rust eventually)?

scripts/redcap_dict_to_properties.py
joelostblom: Hmm I don't think so, it seems like the `v4` suffix in the column name was just to represent the "baseline", which I interpreted as week 0 and visit 4. See https://github.com/onlimit-study/feasibility-data/pull/90/changes#r3310224431 for what I based this on.

But I'm not certain, maybe @K-Beicher can clarify further?

scripts/redcap_dict_to_properties.py
joelostblom: Yes, I think you're right and it seems that week is technically redundant with the visit number already being encoded in the event value. But maybe still a friendlier way of showing it than extracting it from the event and mapping the visit to the corresponding week?

martonvago (dismissed):
I approve pending any input from Kris ☺️

lwjohnst86:
Hm, @joelostblom it closed because it was stacked on a branch that was deleted after merging. Can you fix?

lwjohnst86:
Trying to reopen.

joelostblom:
To reopen, you need to go to the PR you merged and click "Restore branch". Then you can come back here and click "Reopen PR". It's hidden and hard to find out about. I did that now, changed the base branch, and resolved any conflicts.

I contacted github about stacked PRs closing and they said it is a bug and the intended behavior is that the branch should switch to the branch the parent is merged into.

lwjohnst86 (commented):
It looks good to me now, but we just need @K-Beicher to review it and answer the questions.

K-Beicher:
Apologies - this slipped past me when I came back from holiday.

Taking that naming first, technically the ubusy is the correct one, it is short for 'when you are busy', but I agree that it looks better with just busy, and since there isn't a field about others being busy it should be fine. 

I agree that we can remove the information about visit (v) and week numbering from the table names, as long as we keep both in the data, so a column for week_number, and a column for visit_number. The reason being that some visits are in the same week, (technically visits 1-4 are all in week 0), having both will cut down on the mental calculations that either the user or I will have to do if we keep both.

On second thought, it's sat with me for so long that I should probably explicitly tag you all... @joelostblom @martonvago @lwjohnst86

joelostblom:
Thanks @K-Beicher! It sounds like this is ready to merge then @lwjohnst86

scripts/redcap_dict_to_properties.py
```diff
@@ -160,6 +249,21 @@ def _form_to_resource(
...
```
```suggestion
            name="week_number",
            title="Week number of the year",
```

scripts/redcap_dict_to_properties.py
```diff
@@ -160,6 +249,21 @@ def _form_to_resource(
...
+            type="integer",
+            description="The study week when the SEFNC measurement was recorded.",
```
```suggestion
            description="The week number (0-52) in the year when the SEFNC measurement was recorded.",
```

scripts/redcap_dict_to_properties.py
```diff
@@ -160,6 +249,21 @@ def _form_to_resource(
...
+            categories=SEFNC_WEEKS,
+            constraints=sp.ConstraintsProperties(
+                required=True,
+                enum=SEFNC_WEEKS,
+            ),
+        )
+        default_fields.append(week_field)
+        primary_key.append("week")
```
```suggestion
        primary_key.append("week_number")
```

scripts/redcap_dict_to_properties.py
```diff
@@ -216,6 +323,12 @@ def _get_resource_description(form_name: str) -> str:
             "relative to the meal."
         )
 
+    if form_name == "sefnc":
+        return (
+            "Self-efficacy for nutrition change measurements self-reported by "
+            "participants across study weeks."
```
```suggestion
            "Self-efficacy measurements for nutrition change that were self-reported by "
            "participants during the study across the weeks."
```

Just to make it more descriptive and clearer.

scripts/redcap_dict_to_properties.py
lwjohnst86: Related to our talk this morning about primary keys etc, considering the other resources don't all seem to have this week number, the primary key here should be visit number and participant id. But also keep this week number. You'll have to create a visit number column like you did with the week number.

lwjohnst86 (changes requested):
Based on our conversation about primary keys, you'll have to update this to include visit number as a primary key ☺️

scripts/redcap_dict_to_properties.py
martonvago: And what's the relationship between `event` (name) and visit number?

lwjohnst86: I don't know what `event` looks like, so I can't say. But we need some concordance between the columns and instruments ending in `_v[0-9]+` to effectively pivot and connect things.

scripts/redcap_dict_to_properties.py
joelostblom: What confusion are we trying to avoid with this change? To me, "week" could only  be a number. I understand that we might want to clarify what that number represent, e.g. "Week of the year" vs "Week of the study".

With that background, I would suggest something like this:

```suggestion
            name="calendar_week",
            title="Week of the year",
```

Possibly changing `week` to either `calendar_week` or `week_of_the_year`, but I consider just `week` to be sufficient since I don't think the column name needs to capture the full title content.

scripts/redcap_dict_to_properties.py
joelostblom: I attempted this in b18e962. Note that for both week and visit, I've kept them as categories restricted to only the values seen in this dataset. If they can take on other values in other data sets, we could consider changing the to allow for any integer 1 to max_{visit,week}, which is also more true to what they are actually representing even if not all those values are present here. Just raising it for consideration, not sure what is preferable.

scripts/redcap_dict_to_properties.py
lwjohnst86: The confusion is what you stated, "week study" vs "week year". Especially since a lot of ON-LiMiTs presentations and figures show "weeks from start". So this will avoid  that confusion. I like `calendar_week`, that should be clear enough.

scripts/redcap_dict_to_properties.py
```diff
@@ -160,6 +255,33 @@ def _form_to_resource(
         default_fields.append(time_field)
         primary_key.append("minutes_from_meal")
 
+    if form_name == "sefnc":
+        visit_field = sp.FieldProperties(
+            name="visit",
```
```suggestion
            name="visit_id",
```

To follow the pattern "primary keys end in `_id`"

scripts/redcap_dict_to_properties.py
```diff
@@ -160,6 +255,33 @@ def _form_to_resource(
...
+            title="Visit",
```
```suggestion
            title="Visit ID",
```

scripts/redcap_dict_to_properties.py
```diff
@@ -160,6 +255,33 @@ def _form_to_resource(
...
+            type="integer",
+            description="The study visit when the SEFNC measurement was recorded.",
```
```suggestion
            description="The study visit ID for when participants had the SEFNC measurement recorded.",
```

scripts/redcap_dict_to_properties.py
```diff
@@ -160,6 +255,33 @@ def _form_to_resource(
...
+            description="The study week when the SEFNC measurement was recorded.",
+            categories=SEFNC_WEEKS,
+            constraints=sp.ConstraintsProperties(
+                required=True,
+                enum=SEFNC_WEEKS,
+            ),
+        )
+        default_fields.append(visit_field)
+        default_fields.append(week_field)
+        primary_key.append("visit")
```
```suggestion
        primary_key.append("visit_id")
```

lwjohnst86 (changes requested):

scripts/redcap_dict_to_properties.py
joelostblom: Good let's go with `calendar_with` to disambiguate those more clearly.

joelostblom: I thought a bit more about this. Why is it that we think this represents the week of the year? That would imply that they started with all participants the first week of Jan and then followed up in the last week of Dec (week 52), which seems unlikely.

Isn't this column more likely to represent the week of the study? That also seems to align with that they refer to _v4 as baseline and then the other visits as a number of weeks from that baseline. I also found a new place where this relative formulation is used; it's in the `field_label` of the `data_dictionary.json`, e.g. "Tid (i uger) fra baseline: 12<"


```json
{
    "field_name": "book_v6",
    "form_name": "bookinger",
    "section_header": "",
    "field_type": "text",
    "field_label": "<div class=\"rich-text-field-label\"><p>Besøg 6<br><span style=\"font-weight: normal;\"><em>Testdag. </em></span><br><span style=\"font-weight: normal;\"><em>Varighed: 6,5 timer.</em></span><br><span style=\"font-weight: normal;\"><em>Tid (i uger) fra baseline: 12</em></span><br><span style=\"font-weight: normal;\"><em>Dage mellem besøg: SKAL ligge mindst 10 dage efter besøg 5.</em></span></p> <p><em>Husk at booke følgende ressourcer i SP: lokale, DXA, fibro, fødevarepræferencetest. </em></p></div>",
    "select_choices_or_calculations": "",
    "field_note": "",
    "text_validation_type_or_show_slider_number": "datetime_dmy",
    "text_validation_min": "",
    "text_validation_max": "",
    "identifier": "",
    "branching_logic": "",
    "required_field": "y",
    "custom_alignment": "",
    "question_number": "",
    "matrix_group_name": "",
    "matrix_ranking": "",
    "field_annotation": "Booking of visit 6. Workflow."
  }
```

joelostblom: A new complication this brings is that [`v4` is listed as -1 weeks from baseline](https://github.com/onlimit-study/feasibility-data/blob/b7ad11a1586d92c9837c9e5a3165471733db750e/scripts/data_dictionary.json#L1207) (rather than 0): `Tid (i uger) fra baseline: -1` So that suggests that the range should be -1 to 52. 

But everywhere else in `data_dictionary.json`, they refer to v4 as baseline (week 0): `Baseline visit (V4).` and `Uge 0 (baseline), 18, 24, 26 og 51.` and many more.

So I think what this week field represents is the number of weeks after the baseline week. So the first week is this week0 baseline, then the second week is week1 (ie "one after baseline") and so on.

@K-Beicher @lwjohnst86 What do you think?

lwjohnst86 (approved):
Fantastic 😁

scripts/redcap_dict_to_properties.py
K-Beicher: Yes, baseline is often (almost always) week 0, this makes w4 the fourth week after baseline, which means that anything that happens in week four happens four weeks after the participant started on the study.

PR #93, opened 2026-05-27
refactor: ♻️ turn `foer_besoegsdag` into a single resource

### LINKED ISSUE #54: Convert `foer_besoegsdag_*` resources into one
Like with some of the other issues e.g. #46, this should be one resource with a column for `visit` or something similar.

joelostblom:
# Description

This PR combines got a little bit more messy than the other ones because some fields were not in part of each visit and therefore some of them changed from being required to being optional.

But overall this PR does the following:

- Combine `foer_besoegsdag_2`, `foer_besoegsdag_3`, and `foer_besoegsdag_4` into one `foer_besoegsdag` resource.
- Add a required `visit` field with values `2`, `3`, and `4`, and include it in the primary key with `event`.
- Normalize before-visit workflow field names by removing `_wfv2`, `_wfv3`, and `_wfv4` suffixes.
- Mark fields as optional when they are not present across all joined visits.
- Remove visit-specific annotation text such as `Visit 2`, `Visit 3`, `Visit 4`, and `Wfv2`/`Wfv3`/`Wfv4`  when it only repeats the new `visit` field.

Closes #54 

## Checklist

- [x] Ran `just run-all`


martonvago (dismissed):
This makes sense to me in general! 
It's a lot of code, but hopefully it will be possible to generalise some of it 🤞 .

lwjohnst86:
This should be re-opened to main, not stacked. Please avoid stacking as much as possible.

joelostblom:
I do avoid stacking when I can, but use it when it helps the review process (ie avoid looking at a giant irrelevant diff). In this case, I used it because I was trying to generalize some functionality that would be shared between these PRs, but after deciding steer away from that direction, it would have been better if I removed some of the functionality from here and then unstacked these PRs.

No need to close and re-open this though, we can just change the base branch; I just did that and resolved conflicts.

lwjohnst86:
I converted this to draft until #90 has been approved and merged, otherwise it's difficult to know what has been added and what to review.

scripts/redcap_dict_to_properties.py
```diff
@@ -13,9 +13,10 @@
 VAS_TIMEPOINTS = [-10, 30, 60, 90, 120, 180, 240]
 VAS_TIME_FORM_PATTERN = re.compile(r"^vas_(minus10|(30|60|90|120|180|240)_?min)$")
 VAS_TIME_FIELD_PATTERN = re.compile(r"(_fasted)?_(minus10|30|60|90|120|180|240)min$")
-SEFNC_WEEKS = [1, 12, 52]
+
+SEFNC_WEEKS = [0, 12, 52]
 SEFNC_FORM_WEEKS = {
-    "sefnc_baseline_v4": 1,
+    "sefnc_baseline_v4": 0,
```
joelostblom, line 19: I forgot to change this back in the other PR that was just merged. Since this is study week rather than calendar week and the baseline week is included, the range should start at 0

joelostblom (commented):
I think this is ready for another look now that #90 has been merged.

scripts/redcap_dict_to_properties.py
```diff
@@ -211,6 +222,113 @@ def _remove_sefnc_week_from_annotation(annotation: str) -> str:
...
+    """Combines before-visit-day workflow forms into one resource schema."""
+    common_fields = _get_common_foer_besoegsdag_field_names(redcap_fields)
+    return _deduplicate_foer_besoegsdag_fields(
+        _map(
+            redcap_fields,
+            lambda field: _normalise_foer_besoegsdag_visit_resource_field(
+                field, common_fields
+            ),
+        )
+    )
```
lwjohnst86, line 237: Can you revise this and later code to not nest so much? If there were a chain-ability in Python, this could be chained, but it's hard to follow this (and code below) what's actually being output at each step when it's nested like this. ☺️

scripts/redcap_dict_to_properties.py
```diff
@@ -211,6 +222,113 @@ def _remove_sefnc_week_from_annotation(annotation: str) -> str:
...
+
+
+def _get_common_foer_besoegsdag_field_names(
+    redcap_fields: list[dict[str, str]],
+) -> set[str]:
+    visit_fields = _map(
+        _filter(redcap_fields, _is_foer_besoegsdag_visit_resource_field),
```
lwjohnst86, line 244: Same as comment above.

scripts/redcap_dict_to_properties.py
```diff
@@ -211,6 +222,113 @@ def _remove_sefnc_week_from_annotation(annotation: str) -> str:
...
+        lambda field: (
+            _normalise_foer_besoegsdag_field_name(field["field_name"]),
+            FOER_BESOEGSDAG_FORM_VISITS[field["form_name"]],
+        ),
+    )
+    field_visits = reduce(_add_foer_besoegsdag_field_visit, visit_fields, {})
+
+    return set(
+        _map(
+            _filter(
```
lwjohnst86, line 254: Same as comment above.

lwjohnst86 (changes requested):
I started reviewing this but realized we need to take a big step back and rethink how we organise all the files and code, because this file is starting to get difficult to keep track of. But for now, I made a few comments about the flow for the code that could help me review it better.

lwjohnst86 (commented):
As per @joelostblom suggestion about wrangling the data first into `staging` and only after extracting and creating the metadata. I'll convert this to draft for now until we get to that point.

joelostblom:
Thanks! I just created an issue to track this discussion https://github.com/onlimit-study/feasibility-data/issues/116

PR #94, opened 2026-05-28
feat: :sparkles: raw data to staging

### LINKED ISSUE #73: Raw data to staging
> Write scripts to process all those data into `staging/redcap/<resource-name>/<timestamp>.parquet`, where we process the instruments into the correct resource location.

So here we take the unsplit data and split it into resources doing all the resource transformations outlined in the other issues?

martonvago:
# Description

This PR transforms raw data to staged data. For now, this includes special transformations only for VAS.

Closes #73 

## Checklist

- [x] Ran `just run-all`


scripts/stage_data.py
```diff
@@ -0,0 +1,140 @@
...
+from operator import itemgetter
+from pathlib import Path
+from typing import cast
+
+import polars as pl
+import seedcase_soil as so
+
+VAS_TIME_FIELD_PATTERN = re.compile(
+    r"^vas_(?P<field_name>.+?)(_fasted)?_(?P<time>minus10|30|60|90|120|180|240)min$"
+)
```
martonvago, line 12: We can share this with the metadata transformation

scripts/stage_data.py
```diff
@@ -0,0 +1,140 @@
...
+
+def load_raw_data() -> pl.DataFrame:
+    """Loads the latest raw data from `raw/redcap/<timestamp>.csv.gz`."""
+    file_path = Path("raw") / "redcap"
+    files = list(file_path.glob("*.csv.gz"))
+    if not files:
+        raise FileNotFoundError(
+            f"No raw data files found in '{file_path}'. "
+            "Have you run `just download-data`?"
+        )
```
martonvago, line 23: Alternatively, we can just do nothing without throwing an error.

scripts/stage_data.py
```diff
@@ -0,0 +1,140 @@
...
+def _select_with_base_cols(
+    raw_df: pl.DataFrame, resource_name: str, cols: list[str]
+) -> pl.DataFrame:
+    """Selects columns and adds base columns common to all dataframes."""
+    return (
+        raw_df.select(["redcap_event_name"] + cols)
+        .rename({"redcap_event_name": "event"})
+        .with_columns(
+            pl.lit("Copenhagen").alias("center"),
+            pl.lit(resource_name).alias("resource_name"),
```
martonvago, line 53: I'll use this to write the parquet file then drop it

scripts/stage_data.py
```diff
@@ -0,0 +1,140 @@
...
+
+    cols_grouped_by_time: dict[int, list[str]] = {}
+    for col in vas_cols:
+        match = cast(re.Match[str], VAS_TIME_FIELD_PATTERN.match(col))
+
+        time = match.group("time")
+        if time == "minus10":
+            time = "-10"
+
+        cols_grouped_by_time.setdefault(int(time), []).append(col)
```
martonvago, line 100: I can rewrite this dictionary construction without a for loop if you want, but I think it will just be longer and more complicated

.typos.toml
martonvago, line 1: This was done automatically, not sure if it's the right way

.gitattributes
```diff
@@ -1 +1,2 @@
 raw/** filter=lfs diff=lfs merge=lfs -text
+staging/** filter=lfs diff=lfs merge=lfs -text
```
martonvago, line 2: Or we can track `*.parquet`

scripts/stage_data.py
```diff
@@ -0,0 +1,140 @@
...
+    )
+
+
+def _select_with_base_cols(
+    raw_df: pl.DataFrame, resource_name: str, cols: list[str]
+) -> pl.DataFrame:
+    """Selects columns and adds base columns common to all dataframes."""
+    return (
+        raw_df.select(["redcap_event_name"] + cols)
+        .rename({"redcap_event_name": "event"})
```
martonvago, line 50: Having looked at more of the data, I don't think `event` by itself can be the PK. Raised #95

scripts/stage_data.py
lwjohnst86: Yea, I think these types of errors might not be necessary. And might be solved by using some type of "orchestrator", but that's for later. This is fine for now.

scripts/stage_data.py
lwjohnst86: Yea, I was thinking the same.

.gitattributes
lwjohnst86: Not sure, is that a question?

.typos.toml
lwjohnst86: 👍

scripts/stage_data.py
```diff
@@ -0,0 +1,140 @@
...
+        .with_columns(
+            pl.lit("Copenhagen").alias("center"),
+            pl.lit(resource_name).alias("resource_name"),
+        )
+    )
+
+
+def raw_to_staged(raw_df: pl.DataFrame) -> list[pl.DataFrame]:
```
lwjohnst86, line 58: Maybe organize so either all the `_fn` are above or below the normal functions?

scripts/stage_data.py
lwjohnst86, line 53:
```suggestion
            # Only used for creating the Parquet files.
            pl.lit("Copenhagen").alias("center"),
            pl.lit(resource_name).alias("resource_name"),
```

scripts/stage_data.py
```diff
@@ -0,0 +1,140 @@
...
+
+    return _select_with_base_cols(raw_df, resource_name, field_names)
+
+
+def _create_vas_df(raw_df: pl.DataFrame) -> pl.DataFrame:
+    """Creates a dataframe for the VAS resource."""
+    vas_cols = so.keep(
+        raw_df.columns,
+        lambda column: VAS_TIME_FIELD_PATTERN.match(column) is not None,
+    )
```
lwjohnst86, line 90: This would be better by using Polars rather than a filter. E.g. `select()` can take a pattern/exclude

scripts/stage_data.py
```diff
@@ -0,0 +1,140 @@
...
+        time = match.group("time")
+        if time == "minus10":
+            time = "-10"
+
+        cols_grouped_by_time.setdefault(int(time), []).append(col)
+
+    vas_dfs = so.pairwise_fmap(
+        list(cols_grouped_by_time.items()), [raw_df], _create_df_for_time_group
+    )
+    return pl.concat(vas_dfs, how="vertical")
```
lwjohnst86, line 105: I think all of this would be better with a pivot https://docs.pola.rs/user-guide/transformations/pivot/

lwjohnst86 (changes requested):

scripts/stage_data.py
```diff
@@ -0,0 +1,140 @@
...
+
+import polars as pl
+import seedcase_soil as so
+
+VAS_TIME_FIELD_PATTERN = re.compile(
+    r"^vas_(?P<field_name>.+?)(_fasted)?_(?P<time>minus10|30|60|90|120|180|240)min$"
+)
+
+
+def load_raw_data() -> pl.DataFrame:
```
lwjohnst86, line 15: This is where we need to think about the design of the Python files, etc, as we will eventually have non-REDCap raw data, while this only loads the REDCap data. Plus this function name implies reading in all raw data, but it's only reading in the latest. 

Maybe to start, refactor this to have a path as an arg? And I'm not sure we need to only read the latest all the time, we'll want the orchestrater to handle which files to read and which to not read.

lwjohnst86 (changes requested):
Another comment here

scripts/stage_data.py
martonvago: Hmm, yeah, everything here is just about the REDCap data, so like one step in an orchestrated flow. Maybe I will rename things to make that clear, rather than trying to anticipate how it will fit into the flow.

Right now, every time we download the raw REDCap data, we download the full set, that's why I thought taking the latest one made sense. This doesn't quite align with the earlier batch update model, and I was wondering if you had a broader vision for how raw data should be staged. E.g., when would we not stage the latest data?

.gitattributes
martonvago: Just an alternative

martonvago:
I'll park this one for now in favour of #108, which only has simple repeating and non-repeating forms (that need no transformations). It will be easier to build on that than to sort out both the basics and the transformations in one.

PR #108, opened 2026-06-15
feat: :sparkles: stage basic raw data

### LINKED ISSUE #109: Stage forms that need no special transformations
Both repeating and non-repeating forms.

martonvago:
# Description

This PR stages raw data that does not require special transformations (like, e.g., VAS data).
I split this into a separate PR because we are already handling repeating and non-repeating resources, so that's enough complexity.

This works only with #106.
It ran fine on the newest version of the real data and metadata on GenomeDK.

Closes #109 

## Checklist

- [x] Ran `just run-all`


scripts/form_event.json
martonvago, line 1: This mapping tells us what forms/resources are filled in at which events. We can use it to select relevant rows for each resource when staging data.

scripts/redcap_metadata.py
martonvago, line 1: Similar addition as in #106. It might make sense to pull out common logic once both are merged.

lwjohnst86 (commented):
As I've written in other PRs, there's not enough clarity about what data gets converted into what resource. I think we need to take a step back and write some docs to make that link clearer. Plus, as @joelostblom suggested, wrangle data first *without* using/depending on the data dictionary, and only after we've created them into `staging/` will we extract the properties from them and link them to the data dictionary. I'm going to convert this to draft for now.

Issue #98, opened 2026-06-01
Reorganise Python files into modules/folders, for metadata vs data and by resource (not necessarily both)


lwjohnst86:
A nice rolemodel here is how dbt structures their projects: https://docs.getdbt.com/docs/build/projects?version=2.0 dbt is a data warehouse tool heavily used by data engineers for commercial data.

lwjohnst86:
Mostly done now.

PR #112, opened 2026-06-16
docs: 📝 add guide on writing code in `src/`

lwjohnst86:
# Description

It's already getting a bit too disorganized, so this guide explains how the files and folders are organised and used within `src/`.

Related to #98


martonvago (approved):
Good (much needed) plan!

PR #126, opened 2026-06-25
chore: :truck: convert to Python package `src/` layout

### LINKED ISSUE #96: Convert repo into Python package structure, rather than individual scripts with `__main__`
It's starting to get messy, so we need some better structure here. Converting to a package structure will make it a bit easier to organize and use code in our future orchestration tool

lwjohnst86:
# Description

Switch to the `src/` layout. I didn't touch anything else, just a rename.

Closes #96

This PR needs a quick review.


martonvago (changes requested):
I think there are other references to `scripts` that need to be updated (e.g. in `main.py`). Also, should `main.py` be moved as well?

lwjohnst86:
@martonvago when we move to using pytask, we won't need `main.py` anymore. There's several things in Sprout that need to be updated related to this :grimacing: Plus, `main.py` should be outside of the package structure as it is a script (with `__main__`).

martonvago:
@lwjohnst86 okay, that makes sense! But should this PR be in a state where we can run `just run-all` before merging?

lwjohnst86:
Unfortunately, that requires substantially more fixes and `run-all` now includes things like mypy, which now fails very hard :stuck_out_tongue:

lwjohnst86:
It's also not a version bump, so :+1:

martonvago (approved):
Okay 👍 then let's merge it 🚀

Issue #97, opened 2026-06-01
Set up pytask workflow to manage data pipeline


lwjohnst86:
I've made a decision post about this. Use pytask to manage the workflow from raw data to resource (for both data and metadata).

martonvago:
@lwjohnst86 About the data download step: the first implementation downloaded the data and then used the most recent download as input to the other steps. There was some [discussion](https://github.com/onlimit-study/feasibility-data/pull/94/changes#r3340925476) about how the orchestrator should decide which data file to use. Do you have a more concrete idea of how it should choose if not based on which is the latest? 

Or should we keep the data download step separate from the pytask flow and have the pytask flow take the path or file name as input?

martonvago:
We now have the first step of the workflow, so I'm closing this.

PR #135, opened 2026-07-01
docs: 📝 describe flow of REDCap data to staging

lwjohnst86:
# Description

Adds a "flow of data" section, starting with the REDCap flow.

This PR needs a quick review.

## Checklist

- [x] Ran `just run-all`


lwjohnst86:
@martonvago this should answer your question in #97

CONTRIBUTING.md
```diff
@@ -1,5 +1,26 @@
 # Contributing
 
+## Flow of data
+
+### REDCap
+
+The data flows directly from the REDCap API into `raw/redcap/` as a JSON file
```
martonvago: Did you mean csv?

CONTRIBUTING.md
```diff
@@ -1,5 +1,26 @@
...
+with a timestamp appended to the filename. Every time the data is pulled from
+REDCap, a new JSON file is created in `raw/redcap/` with the current timestamp.
+
+Using code writing in `src/feasibility_data/data/redcap/<resource>.py`, each raw
+JSON file is processed into `staging/redcap/<resource>/<timestamp>.parquet`.
+There should be a 1-to-1 mapping between the raw JSON's timestamp and the staged
+resource Parquet file's timestamp.
```
martonvago: Ah, so we always process each data file? That means we should declare separate pytask tasks for the processing of each of them so that we don't recompute old runs when a new data file is added.

One thing I'm wondering about: do we expect REDCap metadata to change over time (it's happened already since we first pulled it)? Because then we would be running some of the the pipeline on old data + updated data dict, which could potentially fail.

CONTRIBUTING.md
```diff
@@ -1,5 +1,26 @@
...
 
+## Flow of data
+
+### REDCap
+
+The data flows directly from the REDCap API into `raw/redcap/` as a JSON file
+with a timestamp appended to the filename. Every time the data is pulled from
+REDCap, a new JSON file is created in `raw/redcap/` with the current timestamp.
+
+Using code writing in `src/feasibility_data/data/redcap/<resource>.py`, each raw
```
```suggestion
Using code written in `src/feasibility_data/data/redcap/<resource>.py`, each raw
```

CONTRIBUTING.md
```diff
@@ -1,5 +1,26 @@
...
+JSON file is processed into `staging/redcap/<resource>/<timestamp>.parquet`.
```
```suggestion
JSON file is processed into a collection of `staging/redcap/<resource>/<timestamp>.parquet` files.
```
To make it clear that one timestamped data file is processed into multiple timestamped parquet files

martonvago (changes requested):
Thanks for the explanation! Just some questions

CONTRIBUTING.md
lwjohnst86: No, we don't need to create separate pytask tasks for each file. Give a list of paths, read them each in via map. Doesn't need to be complicated. I (hope/think) pytask is smart enough to handle these types of situations (I know the R package targets can do it, so it is possible to do).

About the metadata, ideally we always pull from REDCap and update from their changes. That way we can make sure there aren't any issues even from pulling  the data. If it fails, that's a good sign for us to investigate and debug.

martonvago: Yeah, for sure, that would work, but I think it would also rerun the pipeline on each timestamped data file every time. I don't know if that's okay? As far as I understand, pytask's solution to this is looping over the inputs and treating each one as a task ([like so](https://pytask-dev.readthedocs.io/en/stable/tutorials/repeating_tasks_with_different_inputs.html) or [so](https://pytask-dev.readthedocs.io/en/stable/how_to_guides/provisional_nodes_and_task_generators.html)).

The problem with the REDCap metadata is that we only keep one version: the latest one we just pulled. The data processing pipeline takes as input a timestamped data file and the REDCap metadata. If we run the pipeline on older timestamped data files, there is no guarantee they won't fail with the current REDCap metadata (which is the only one we have). This is why I got hung up on rerunning the pipeline on older data files.

lwjohnst86: I expanded on the explanation to answer your questions.

martonvago (approved):
Thanks, that clears it up!

PR #138, opened 2026-07-01
docs: 📝 explain how to use `core.py` files

lwjohnst86:
# Description

This adds to the layout description of the `src/` Python files.

Needs a quick review.

## Checklist

- [x] Ran `just run-all`


CONTRIBUTING.md
```diff
@@ -20,6 +20,12 @@ under `src/feasibility_data/` is:
   first before being processed into the final data resource as Sprout needs to
   run checks against the metadata before converting it into the final data
   resource.
+- In either the `data/` or `metadata/` directories, files named `*/core.py`
+  contain functions that do general or common processing tasks related to the
+  parent folder name. For example, `metadata/core.py` contains functions for
+  top-level metadata processing that isn't specific any one source or resource,
```
```suggestion
  top-level metadata processing that isn't specific to any one source or resource,
```

CONTRIBUTING.md
```diff
@@ -20,6 +20,12 @@ under `src/feasibility_data/` is:
...
+  while `metadata/redcap/core.py` contains functions for processing REDCap data
```
```suggestion
  while `data/redcap/core.py` contains functions for processing REDCap data
```
?

CONTRIBUTING.md
```diff
@@ -20,6 +20,12 @@ under `src/feasibility_data/` is:
...
+  that isn't specific to any one resource.
```
```suggestion
  that aren't specific to any one resource.
```

CONTRIBUTING.md
martonvago, line 1: The general idea makes sense, just a question about how this will be used in practice (based on the PRs that start implementing this structure). Do we not want to structure the code more around pytask tasks? I mean annotating a function as a task right where it is defined instead of importing it in `build.py` just to annotate it there. Then each task could live in its own file inside the appropriate folder (as described here).

martonvago, line 1: Do we want to align the common vs core naming? Feels like the `common` folder and the `core.py` files do a similar kind of organising.

martonvago (changes requested):
Nice, just some questions ☺️

lwjohnst86:
@martonvago I added more content to address your questions :relaxed:

CONTRIBUTING.md
```diff
@@ -20,10 +20,34 @@ under `src/feasibility_data/` is:
...
+  metadata processing that isn't specific to any one source or resource, while
+  `data/redcap/core.py` contains functions for processing REDCap data that
+  aren't specific to any one resource. This `core.py` file can be treated like
+  the `__init__.py` file. We don't use `__init__.py` files to store functions as
+  the semantic meaning of `__init__.py` is to initialise the folder as part of
+  the package. The semantic meaning of `core.py` is to be a collection of
+  functions that are used in its parent source/resource folder.
+- `common/`: Contains functions that are used across *all* (or many) Python
+  files, between metadata and data or between sources/resources. This is not the
+  same as the `**/core.py` file that are *specific* to the particular source or
```
```suggestion
  same as the `**/core.py` files that are *specific* to the particular source or
```

martonvago: I think I understand what you mean, but it may be a bit confusing that this paragraph says `core.py` is specific to a particular source or resource, but the previous paragraph says `core.py` contains functions that aren't specific to any source or resource.

martonvago (changes requested):

CONTRIBUTING.md
lwjohnst86: Oops, that was a typo.

lwjohnst86: I fixed it, hopefully that's clearer.

CONTRIBUTING.md
```diff
@@ -20,10 +20,35 @@ under `src/feasibility_data/` is:
...
   resource.
-- `common/`: Contains functions that are used across multiple resources. The
-  names of the Python files within are not standardized, but they should be
-  descriptive of the overall functionality they provide within.
+- In either the `data/` or `metadata/` directories, files named `*/core.py`
+  contain functions that do general processing tasks related to the parent
+  folder name. For example, `metadata/core.py` contains functions for top-level
+  metadata processing that is for general metadata, but not strictly tied to any
+  given source or resource, such as data package-level metadata. Meanwhile,
+  `data/redcap/vas/core.py` contains functions for processing REDCap data that
```
martonvago: Just to make sure I understand, what's the difference between `data/redcap/vas.py` (logic for processing VAS data) and `data/redcap/vas/core.py` (general VAS data processing logic)?

lwjohnst86: Dammit, another mistake :facepalm: :zany_face:  I'll fix that.

martonvago (approved):
Okay, I think I finally get it ahahaha 😁

Issue #132, opened 2026-06-26
Consider reorganizing shared code


joelostblom:
> I started reviewing this but realized we need to take a big step back and rethink how we organise all the files and code, because this file is starting to get difficult to keep track of.

_Originally posted by @lwjohnst86 in https://github.com/onlimit-study/feasibility-data/pull/93#pullrequestreview-4504338702_

Depending on what we decide for https://github.com/onlimit-study/feasibility-data/issues/116 this might not be as relevant anymore

lwjohnst86:
This should be done now, with all of the updates, etc.

PR #141, opened 2026-07-01
chore: 🚚 move Python code to write dictionary into `core.py`

lwjohnst86:
# Description

To start following the new structure.

Related #132 

Needs a quick review.

## Checklist

- [x] Ran `just run-all`


src/feasibility_data/metadata/redcap/core.py
```diff
@@ -8,7 +8,19 @@
 load_dotenv()
 
 
-def get_data_dict_from_redcap() -> list[dict[str, str]]:
+def write_dictionary() -> Path:
+    """Writes the data dictionary to `src/feasibility_data/metadata/redcap/dictionary.json`."""
```
martonvago: Do we eventually want to store this in a pytask-style `bld` folder?

src/feasibility_data/metadata/redcap/core.py
martonvago, line 1: `CONTRIBUTING.md` says that code that creates the final metadata lives in `metadata` and code that creates the final data lives in `data`, with shared code in `common`. The REDCap data dict is used for creating both the final metadata and the final data, so I was wondering if it should live in `common` instead?

martonvago, line 1: Will this eventually be one task (e.g. `download_data_dictionary`)? That is how I've been thinking of it / writing it at least.

lwjohnst86: No, common is for functions. `dictionary.json` is metadata, so should be in metadata. But, based on your lower comment, I think you're right that it should eventually be kept in `bld/` 👍 But we can do that once the pytasks are set up ☺️

src/feasibility_data/metadata/redcap/core.py
lwjohnst86: Yea, that makes sense 👍 I'll modify this a bit to match that (future) plan.

src/feasibility_data/metadata/redcap/core.py
lwjohnst86: Yea exactly 👍

martonvago: Okay! And what about the function downloading the data dict?

lwjohnst86:
@martonvago I updated the PR to include the future pytask integration plus for when we have dictionaries from each center.

martonvago (approved):
Nice, thanks, I will put the pytask things on top of this then!

PR #144, opened 2026-07-01
docs: 📝 add logos to bottom of landing page

### LINKED ISSUE #133: Include logos in footer
Only one the landing page though.

lwjohnst86:
# Description

Closes #133

Needs a quick review.

## Checklist

- [x] Ran `just run-all`


signekb (changes requested):
The height doesn't seem to do what we want here, resulting of logos with quite different sizes.

Wide/normal screen: 

Image

Narrow screen:
 

Image

You could add the logos to the footer here instead as on the main website? 

https://github.com/onlimit-study/main-website/blob/be549bbc505b3478faad5046c836be5348ecab75/_quarto.yml#L29

lwjohnst86:
I can't seem to be able to put footers on individual pages (which makes sense). And yea, I don't love the sizing issue, but I think that's because of the original images themselves being different sized. I don't know an easy fix for that without spending too much time on it.

lwjohnst86:
@signekb ah, I can switch to width, not height :relaxed:

signekb (approved):
Nice(r)! It does go from 3 cols to 1 on narrow screens (where there seems to still be room for 2 cols), but this is fine 🌷

PR #153, opened 2026-07-02
chore: :recycle: add download metadata to pytask pipeline

martonvago:
# Description

This PR sets up pytask and downloads the data dict from REDCap. As we will be downloading other metadata as well, I'm calling it field metadata because it contains metadata about all fields in REDCap.

Related to #97 

Needs a thorough review.

## Checklist

- [ ] Ran `just run-all`


src/feasibility_data/common/redcap.py
```diff
@@ -0,0 +1,60 @@
...
+class Center(Enum):
+    """The centers in the study."""
+
+    # TODO: Update these once we can connect them.
+    Aarhus = APIConfig(env_key="REDC_AAR_API_KEY", url="https://redcap.auh.dk/api/")
+    Copenhagen = APIConfig(
+        env_key="REDC_CPH_API_KEY", url="https://redcap.regionh.dk/api/"
+    )
+    Odense = APIConfig(env_key="REDC_ODN_API_KEY", url="https://redcap.sdu.dk/api/")
+    Test = APIConfig(env_key="TEST_API_KEY", url="https://redcap.au.dk/api/")
```
martonvago, line 29: Can this live here for testing?

src/feasibility_data/build.py
martonvago, line 1: If we don't want to keep passing the paths around that declare task dependencies, we can move them into a [data catalog](https://pytask-dev.readthedocs.io/en/stable/tutorials/using_a_data_catalog.html)

src/feasibility_data/build.py
```diff
@@ -0,0 +1,21 @@
...
+BLD = SRC.joinpath("..", "bld").resolve()
+
+BLD_REDCAP = BLD / "redcap"
+
+
+def task_download_field_metadata(
+    field_metadata_path: Annotated[Path, Product] = BLD_REDCAP / "field_metadata.json",
+) -> None:
+    """Download field metadata to `BLD_REDCAP`."""
+    download_redcap_metadata(field_metadata_path, "metadata")
```
martonvago: As we will be downloading other metadata as well, I'm calling this field metadata because it contains metadata about all fields in REDCap.

src/feasibility_data/build.py
martonvago, line 1: Maybe this is a good place to discuss the main issue/question I've run into with pytask so far: download tasks declare that they produce a file or files. If that file is unchanged, they [won't rerun](https://pytask-dev.readthedocs.io/en/stable/how_to_guides/remote_files.html?h=force#detecting-changes-in-remote-files) automatically when the pipeline is run with `uv run pytask`. We can use the `--force` flag on the CLI to get around this. Is that a workable solution or should I look for another way?

src/feasibility_data/common/redcap.py
martonvago, line 1: I also put up the changes for pulling out common logic separately as #154 because I wasn't sure how it was most helpful.

src/feasibility_data/build.py
lwjohnst86: Yes, that behaviour is expected. And yes, we'd have to include some "force" flag for those specific tasks. Or just always use `--force`. I'd have to look through the docs to see if there was an option to always run a task.

src/feasibility_data/build.py
```diff
@@ -0,0 +1,21 @@
...
+)
+
+SRC = Path(str(files("feasibility_data"))).joinpath("..").resolve()
+BLD = SRC.joinpath("..", "bld").resolve()
+
+BLD_REDCAP = BLD / "redcap"
+
+
+def task_download_field_metadata(
+    field_metadata_path: Annotated[Path, Product] = BLD_REDCAP / "field_metadata.json",
```
lwjohnst86: I'm a bit confused. What is `metadata` in this case? The dictionary?

lwjohnst86 (dismissed):

src/feasibility_data/build.py
martonvago: Would you like me to do that now?

src/feasibility_data/build.py
martonvago: Yes: we will be downloading multiple dictionaries from REDCap that all contain metadata (about events, repeating instruments, fields). So I'm calling what REDCap calls "data dictionary" "field metadata" here because it contains metadata about the form fields.

lwjohnst86: What is the reason to switch to using "field metadata" when we were using "data dictionary" before?

src/feasibility_data/build.py
lwjohnst86: Yea, let's see how it looks :relaxed:

src/feasibility_data/build.py
martonvago: Do you mean more than the above?
That is: initially we thought we would only need to make one metadata request to REDCap. This was for the "data dictionary" that contains metadata about the form fields. "data dictionary" is (one of) REDCap's term for this metadata, so we used it as well. Later (while looking into primary keys), we found out that this is not enough and that, in fact, we need 3 pieces of metadata: about fields, events, and repeating instruments. So now that we are building things up from square one, I thought it would be better to name these 3 kinds of metadata consistently and descriptively. 

(I also think that calling the metadata "dictionary" is quite vague because in Python many things are dictionaries. Plus, the data dict is actually a list of dictionaries.)

src/feasibility_data/build.py
martonvago: I had a deeper look and now I think we don't want to use this actually 😅 . 
It looks like the main use case for a data catalog is to decouple input/output objects from their storage location, but, for us, the entire flow is structured around files being at particular locations. Plus, the default "node" they use to represent arbitrary objects stores them in binary, so that's not great for inspecting or version controlling. 

It is possible to use "path nodes" in a data catalog, but that is just a slightly more verbose way of defining a constant for each reused path.

Now, the one thing we could potentially do is define [custom Pytask "nodes"](https://pytask-dev.readthedocs.io/en/stable/how_to_guides/writing_custom_nodes.html) for the file types we work with (CSV, JSON, Parquet). These nodes could encapsulate read/write operations, which is the most cumbersome thing about passing around paths. This would obviously add complexity, so let me know if you think it's worth it.

src/feasibility_data/build.py
```diff
@@ -0,0 +1,23 @@
...
+
+import feasibility_data.common.json as cj
+import feasibility_data.common.redcap as cr
+
+SRC = Path(str(files("feasibility_data"))).joinpath("..").resolve()
+BLD = SRC.joinpath("..", "bld").resolve()
+
+BLD_REDCAP = BLD / "redcap"
+
+FIELD_METADATA_PATH = BLD_REDCAP / "field_metadata.json"
```
martonvago, line 15: If we don't want the constants floating around, we could pack them in a class.

src/feasibility_data/build.py
signekb: To move things forward, I'll make a separate issue and resolve this conversion 🌷  See #162.

signekb: Made an issue to remind us of this #164.

src/feasibility_data/build.py
signekb: Made a discussion issue for this #163

src/feasibility_data/common/redcap.py
```diff
@@ -0,0 +1,60 @@
...
```
signekb, line 29:
```suggestion
    # TODO: Can this live here for testing? 
    Test = APIConfig(env_key="TEST_API_KEY", url="https://redcap.au.dk/api/")
```

src/feasibility_data/build.py
signekb: Also made an issue about this discussion in #165 and will resolve it here to move things forward.

src/feasibility_data/common/redcap.py
```diff
@@ -0,0 +1,60 @@
...
+
+import requests
+from dotenv import load_dotenv
+
+load_dotenv()
+
+
+@dataclass
+class APIConfig:
+    """Configuration for the REDCap API."""
```
signekb, line 14:
```suggestion
    """Configuration for the APIs."""
```
Could this be used for other APIs, e.g., MyFood24? If yes, maybe it should be generalised (basically just removing REDCap from the docstring) and placed in a file that's called something else. E.g., `api.py`? Or maybe that would be overkill if we only expect this one simple class to be common across APIs?

src/feasibility_data/common/redcap.py
signekb, line 1: I'm wondering if some of these REDCap API functions could be generalised to be used for other API calls, such as for downloading MyFood24 data? Or do they contain something too specific to be able to be generalised? 
 (I'm coming from a POV of not knowing much about the data itself and working with APIs)

signekb (changes requested):
This looks good to me :sunflower: To move things forward, I've closed a few conversations and made issues for them. I've added some minor questions and suggestions:

src/feasibility_data/common/redcap.py
signekb: Oh, just saw that Luke said yes to this in #154. Will resolve this.

signekb (approved):
Seeing that my questions and suggestions have been addressed in #154, I'll approve and merge 🚀

src/feasibility_data/build.py
signekb: Will be addressed in #165

PR #154, opened 2026-07-06
chore: ♻️ commonise REDCap API and JSON logic

martonvago:
# Description

This PR pulls out shared logic for using the REDCap API and (relatedly) reading/writing JSON.

This can be treated as groundwork for #153 if it's more helpful to review this first.

Needs a thorough review.

## Checklist

- [x] Ran `just run-all`


src/feasibility_data/common/redcap.py
```diff
@@ -0,0 +1,60 @@
...
+class Center(Enum):
+    """The centers in the study."""
+
+    # TODO: Update these once we can connect them.
+    Aarhus = APIConfig(env_key="REDC_AAR_API_KEY", url="https://redcap.auh.dk/api/")
+    Copenhagen = APIConfig(
+        env_key="REDC_CPH_API_KEY", url="https://redcap.regionh.dk/api/"
+    )
+    Odense = APIConfig(env_key="REDC_ODN_API_KEY", url="https://redcap.sdu.dk/api/")
+    Test = APIConfig(env_key="TEST_API_KEY", url="https://redcap.au.dk/api/")
```
martonvago, line 29: Can this live here?

src/feasibility_data/metadata/redcap/core.py
```diff
@@ -1,67 +1,10 @@
...
-        "token": token,
-        "content": "metadata",
-        "format": "json",
-        "returnFormat": "json",
-    }
-    response = requests.post(center.value.url, data=data, timeout=30)
-    response.raise_for_status()
-    dictionary: list[dict[str, str]] = response.json()
-    return dictionary
+def download_redcap_metadata(path: Path, content: str) -> None:
```
martonvago: Calling this metadata because it will be used to download different types of metadata (event info, field metadata, repeating instruments)

src/feasibility_data/common/redcap.py
lwjohnst86, line 1: For now, we can keep this as `common/redcap.py`.

src/feasibility_data/common/json.py
lwjohnst86, line 1: This isn't strictly just redcap functionality (could even be moved into Soil). So this could be moved into `common/json.py`.

src/feasibility_data/common/redcap.py
lwjohnst86: Fo sho ☺️

src/feasibility_data/common/redcap/api.py
```diff
@@ -0,0 +1,60 @@
...
+    # TODO: Update these once we can connect them.
+    Aarhus = APIConfig(env_key="REDC_AAR_API_KEY", url="https://redcap.auh.dk/api/")
+    Copenhagen = APIConfig(
+        env_key="REDC_CPH_API_KEY", url="https://redcap.regionh.dk/api/"
+    )
+    Odense = APIConfig(env_key="REDC_ODN_API_KEY", url="https://redcap.sdu.dk/api/")
+    Test = APIConfig(env_key="TEST_API_KEY", url="https://redcap.au.dk/api/")
+
+
+def get_from_redcap(
```
lwjohnst86: Since this function is already in the `redcap.py` file (or soon to be), we don't need to use `redcap` in the function name.


```suggestion
def get_request(
```

Will also help make it clearer what `data` arg is if the name is `get_request()` (I was a bit confused at first, because we're processing data for the data package).

src/feasibility_data/data/redcap/raw.py
lwjohnst86, line 1: The functions outside of the pytask build file are a bit like how we've built functions in our other packages. They are small focused functions, that we build up in the pytask function. So here, this doesn't need to be a final "big single function that does everything".

src/feasibility_data/metadata/redcap/core.py
lwjohnst86: If there are only fixed numbers of content types, this `content` should be a new enum/literal type to make those fixed inputs explicit.

lwjohnst86: This can be two steps, not one: get from redcap, write to file.

src/feasibility_data/common/redcap/api.py
```diff
@@ -0,0 +1,60 @@
...
+
+    data["token"] = token
+
+    response = requests.post(center.value.url, data=data, timeout=60)
+    response.raise_for_status()
+
+    return response
+
+
+def get_json_from_redcap(
```
lwjohnst86:
```suggestion
def get_json_request(
```

Related to above.

src/feasibility_data/common/redcap/api.py
```diff
@@ -0,0 +1,60 @@
...
+    content: str,
+    center: Center = Center.Copenhagen,
+) -> Any:
+    """Send a request to the REDCap API and return the JSON response."""
+    data = {
+        "content": content,
+        "format": "json",
+        "returnFormat": "json",
+    }
+    response = get_from_redcap(data, center)
```
lwjohnst86:
```suggestion
    response = get_request(data, center)
```

src/feasibility_data/data/redcap/raw.py
```diff
@@ -1,48 +1,37 @@
...
 from io import StringIO
 from pathlib import Path
 
 import polars as pl
-import requests
-import seedcase_soil as so
-from dotenv import load_dotenv
 
-load_dotenv()
+from feasibility_data.common.redcap.api import Center, get_from_redcap
```
lwjohnst86:
```suggestion
import feasibility_data.common.redcap as cr
```

And then use `cr.` to select things, so that we can keep function names smaller, but use the imports to clearly indicate where they are coming from.

src/feasibility_data/data/redcap/raw.py
```diff
@@ -1,48 +1,37 @@
...
 
 
-def write_raw() -> Path:
-    """Writes the raw data to `raw/redcap/<timestamp>.csv.gz`."""
-    data = _request_raw()
+def download_redcap_data(
```
lwjohnst86: Split this into two smaller functions: get data and write data, so that they can be built up in pytasks build file.

lwjohnst86 (dismissed):
`refactor`, `feat`, `fix` tags are only used when data or metadata are changed in the PR. Since neither have been changed here, this is `chore` or some other tag.

src/feasibility_data/metadata/redcap/core.py
martonvago: Sure! I bundled them because this will be the flow for all metadata downloads, but it's not a big thing to repeat this in those places.

src/feasibility_data/common/redcap.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def get(
```
martonvago, line 32: Called it `get` because `get_request` was a bit confusing because we are sending a POST request and getting a response. But moved to `request_data` as a param name.

signekb (approved):

Issue #55, opened 2026-04-08
Exclude tables administrative tables from data package


lwjohnst86:
Like `gruppemde_uge_1` or `hba1c_uge_30`. These are administrative data, which might be better placed in an admin data package. So these types of resources shouldn't be included in this data package.

K-Beicher:
[DataDictionaryDownload20260702.xlsx](https://github.com/user-attachments/files/29636140/DataDictionaryDownload20260702.xlsx)
@signekb Here is the file I've been working on. there are three suffixes _admin and _file which should be excluded, and _dq which will need to be processed to some level as I need them for data quality checks.

signekb:
It seems that the new field names and current field names are reversed in the file above.

K-Beicher:
New file:

[DataDictionaryDownload20260702.xlsx](https://github.com/user-attachments/files/29636397/DataDictionaryDownload20260702.xlsx)

PR #159, opened 2026-07-08
chore: 🍱 add REDCap data dictionary with field and form renames

signekb:
# Description

@K-Beicher added this in #55. I've converted it to csv, renamed it, and renamed the columns.

Related to #55. 

Needs a quick review. Did we want to use `chore` for most PRs in this repo for now?

## Checklist

- [X] Ran `just run-all`


src/feasibility_data/metadata/redcap/source-dictionary.csv
lwjohnst86, line 1: Damn this is an ugly formatted file! Could you rename to `dictionary.csv` or `source-dictionary.csv`?

lwjohnst86 (dismissed):

src/feasibility_data/metadata/redcap/source-dictionary.csv
signekb: @lwjohnst86 Hmm, maybe we should actually name it in a way that clearly tells what's different about this file: the new field and form names. So, how about something like`dictionary-name-mapping`?

signekb: We could even remove all the unnecessary columns; if we only want to use this for name mapping, we might not need more than the first four cols.

lwjohnst86: That's true. We only the the new and old names. But I think we should still have this original dictionary here, since it was used to create the REDCap structure and questions.

signekb: Oh, I thought it was just downloaded from REDcap - hence the original name in #55 (`DataDictionaryDownload20260702.xlsx`)

signekb: Can you help us here, @K-Beicher ? :)
Edit: 
- Is this dictionary the "original" dictionary used to create the REDCap structure and questions? 
- If not, do we need other columns here than the first four?

K-Beicher: This is a created document that the team uses to create an overview of the fields, it is based on the data dictionary which can be downloaded from REDCap (and also used to manipulate REDCap), which is even worse to read. I think as long as we have the old and new names the rest can go, I've used some of the fields to make sure that I know what is going on, as not all the field names are easy to decipher.

signekb: Okay, I'll keep all the columns, just in case then 🌷

K-Beicher (approved):
Looks good!

PR #161, opened 2026-07-10
docs: 📝 describe release process in CONTRIBUTING

lwjohnst86:
# Description

This is hopefully a starting point for us to figure out how to work together and how releases should happen.

Needs a thorough review.

## Checklist

- [x] Ran `just run-all`


CONTRIBUTING.md
```diff
@@ -43,16 +43,99 @@ terminal so that the working directory is the root of this project
...
 ```
 
+## Release process
+
+Usually the release process for other types of packages is done through a
+`release.yml` GitHub workflow. But for data packages it's a bit different,
+especially when there is human data involved and when the data is on secure
+servers.
+
+- Because the data is on secure servers, we can't use GitHub workflows. Which
```
```suggestion
- Because the data is on secure servers, we can't use GitHub workflows. This
```

CONTRIBUTING.md
```diff
@@ -43,16 +43,99 @@ terminal so that the working directory is the root of this project
...
+Which Conventional Commit type you use depends on the content of the commit
+message. We use aspects of [Data Package's semantic
+versioning](https://datapackage.org/recipes/data-package-version/), where `feat`
+updates the `MINOR` version, `fix` and `refactor` updates the `PATCH` version,
+and any `BREAKING CHANGE` (in the commit message footer) or `<type>!` (e.g.,
+`feat!`) updates the `MAJOR` version. The final format is `MAJOR.MINOR.PATCH`.
+
+<!-- TODO: When should a "stable release" be? After all participants go through the first phases? -->
+
+Breaking changes with the `<type>!` format only happens after the first stable
```
```suggestion
Breaking changes with the `<type>!` format only happen after the first stable
```

CONTRIBUTING.md
```diff
@@ -43,16 +43,99 @@ terminal so that the working directory is the root of this project
...
+  anymore. Instead, a release will only be created on a schedule (if there are
+  any releasable changes).
+
+<!-- TODO: How often should we release? -->
+
+The steps for creating a release are:
+
+- Check the commit history since the last release for any releasable changes. If
+  no releasable changes are found, then no release is created. Otherwise, the
+  process continues.
```
martonvago, line 128: What if there are no changes in the history but there has been a data / metadata update? We would only find that out in step 3 here.

martonvago (changes requested):

CONTRIBUTING.md
```diff
@@ -43,16 +43,99 @@ terminal so that the working directory is the root of this project
...
+- The first release happens once there is code that takes the first resource and
+  its metadata from raw into its final state. The code must also be integrated
+  into the `build.py` file, so that the build pipeline can run automatically.
+  Note that we've already created a few releases, but we realised this process
+  didn't work well for data packages and switched to this scheduled process.
+- Because the release is scheduled, each pull request won't trigger a release
+  anymore. Instead, a release will only be created on a schedule (if there are
+  any releasable changes).
+
+<!-- TODO: How often should we release? -->
```
joelostblom, line 122: Funnily, I implemented something similar in Altair just a few weeks ago. There, we are starting with twice a month on the 1st and 15th. Ideally I would have wanted a certain weekday, e.g. every first and third Mon of the month (so that a release doesn't happen at a time when no one is available to fix issues, but that seemed impossible to configure, at least via GitHub action's cronjob. There, I also added a step that includes human review, so a release is drafted twice a month, but then a human approves it and optionally expands on the release notes with some examples of new functionality if needed. So those are just some considerations to think about if any of them would apply here too.

joelostblom (commented):
Looks good to me overall, just a small comment

CONTRIBUTING.md
lwjohnst86: Nice! Yea, thinking about this, I think we'll start with a manual release via a justfile recipe `just release`. Mainly because the cron job would need some credentials in order to push the tag and commits to the `main` branch.

CONTRIBUTING.md
lwjohnst86: Hmmmm, that is an excellent question 🤔 I'll need to update the build guide here and in `data-pkg-guide`. I think what we'll do is pull the data and commit it and have a PR with only that updated data. And occasionally manually do the build and release, rather than automated, since we don't know if the data update is a `feat`, `fix`, or `refactor`.

lwjohnst86: I'll make those updates in another PR, where I'll also update the corresponding text in `data-pkg-guide` (or trim this down and point to that guide).

martonvago (approved):

PR #157, opened 2026-07-08
docs: 📝 describe the build process in CONTRIBUTING

lwjohnst86:
# Description

Related to the discussion we had during the meeting.

Related to #156

Needs a thorough review.

## Checklist

- [x] Ran `just run-all`

PR #170, opened 2026-07-24
chore: 🍱 add event and repeating forms metadata download to pytask

### LINKED ISSUE #166: Download metadata from REDCap

martonvago:
# Description

This PR adds Pytask tasks for downloading event and repeating forms metadata. We will use these later in the flow.

Closes #166 

Needs a quick review.

## Checklist

- [x] Ran `just run-all` -- with expected failures


.gitignore
```diff
@@ -50,3 +50,6 @@ docs/index.qmd
 
 # Misc files
 *.log
+
+# Pytask products
+bld/
```
martonvago, line 56: We mentioned elsewhere that we don't want to commit intermediary files (so, for metadata, only the final `datapackage.json`), so I added the Pytask `bld` folder to gitignore.

signekb: Oh, I thought we would treat the "raw" metadata downloaded from REDCap as the data and save it to `raw/`? Maybe I've missed a discussion (or misunderstood this)? 

@lwjohnst86 can probably answer this.

signekb (commented):
LGTM. Only a question about where we want to write the metadata files to.

.gitignore
martonvago: Kinda relevant: https://github.com/onlimit-study/feasibility-data/pull/157#discussion_r3578857181

I think the idea was that we [only want to run the build pipeline with the latest metadata](https://github.com/onlimit-study/feasibility-data/blob/4d64958c36e76983cbdb5b462355e2334da2a8af/CONTRIBUTING.md?plain=1#L113), so we don't need a historical record of how it changed. And there is no point saving the latest metadata because we will pull a fresh copy anyway every time the pipeline runs.

src/feasibility_data/build.py
```diff
@@ -13,6 +13,8 @@
 BLD_REDCAP = BLD / "redcap"
 
 FIELD_METADATA_PATH = BLD_REDCAP / "field_metadata.json"
+EVENT_METADATA_PATH = BLD_REDCAP / "event_metadata.json"
+REPEATING_FORMS_METADATA_PATH = BLD_REDCAP / "repeating_forms_metadata.json"
```
lwjohnst86, line 22: I still don't understand what these are. Can you make a PR and add an entry to the contributing file to describe these? I have a suspicion that we could rename these to be clearer to our domain ("event" is so generic and "repeating form" doesn't say what it is repeating).

.gitignore
lwjohnst86: Yea, that's right ☺️ these intermediary files are temporary/only used for specific points within the build, so *theoretically* we don't need them. But we can see how it goes and reassess.

lwjohnst86 (changes requested):

src/feasibility_data/build.py
martonvago: Okay, I'll make a PR!

As for the naming, I'd lean towards keeping it aligned with REDCap terminology because these files will only be used in a REDCap context (when transforming the REDCap data to our representation). The REDCap data comes with columns that reference events and repeating forms/instruments, so it might be less confusing to match these. Then, in our own representation, we can rename them or the columns we derive from them.

martonvago: Done in #181

signekb (approved):

lwjohnst86 (approved):

Issue #163, opened 2026-07-22
[discussion] add class for Pytask constants? 


signekb:
> If we don't want the constants floating around in `src/feasibility_data/build.py`, we could pack them in a class.

_Originally posted by @martonvago in https://github.com/onlimit-study/feasibility-data/pull/153#discussion_r3545016564_

PR #171, opened 2026-07-24
chore: 🏗️ add REDCap data download step to pytask

### LINKED ISSUE #167: Download data from REDCap

martonvago:
# Description

This PR adds the REDCap data download step to Pytask.

Closes #167 

Needs a quick review.

## Checklist

- [x] Ran `just run-all`


src/feasibility_data/build.py
```diff
@@ -2,15 +2,18 @@
 from pathlib import Path
 from typing import Annotated
 
-from pytask import Product
+from pytask import DirectoryNode, Product
 
 import feasibility_data.common.json as cj
 import feasibility_data.common.redcap as cr
+import feasibility_data.data.redcap.raw as dr
```
martonvago, line 9: These abbreviations are starting to get a bit confusing 😅

src/feasibility_data/build.py
```diff
@@ -21,3 +24,16 @@ def task_download_field_metadata(
...
+
+def task_download_data(
+    raw_data_dir: Annotated[
+        Path,
+        DirectoryNode(root_dir=RAW_REDCAP, pattern="*.csv.gz"),
+        Product,
+    ],
+) -> None:
+    """Download the latest data from all centers to `RAW_REDCAP/<timestamp>.csv.gz`."""
+    for center in [cr.Center.Copenhagen]:
```
martonvago, line 38: This is a nod towards being aware that we need to handle different centers without actually handling them. I don't think handling centers has been completely fleshed out yet.

src/feasibility_data/build.py
```diff
@@ -21,3 +24,16 @@ def task_download_field_metadata(
     """Download field metadata to `BLD_REDCAP`."""
     metadata = cr.get_json("metadata")
     cj.write_json(field_metadata_path, metadata)
+
+
+def task_download_data(
+    raw_data_dir: Annotated[
+        Path,
+        DirectoryNode(root_dir=RAW_REDCAP, pattern="*.csv.gz"),
```
martonvago, line 32: Tells Pytask that this task produces files in the specified directory that match the pattern. Later steps will be triggered by changes to these files (i.e. file additions).

src/feasibility_data/data/redcap/raw.py
```diff
@@ -30,7 +30,7 @@ def download_data(center: cr.Center) -> str:
 def write_data(raw_data_dir: Path, data: str) -> None:
     """Write the data as a timestamped file."""
     df = pl.read_csv(StringIO(data), separator=";", infer_schema=False)
-    timestamp = datetime.now().strftime("%Y-%m-%dT%H-%M-%S")
+    timestamp = datetime.now(UTC).strftime("%Y-%m-%dT%H-%M-%SZ")
```
martonvago: Was flagged by Ruff

src/feasibility_data/build.py
signekb: Yeaaa, I agree. Maybe we'll find a nicer system once we get more modules.

src/feasibility_data/build.py
```diff
@@ -2,15 +2,18 @@
...
 import feasibility_data.common.json as cj
 import feasibility_data.common.redcap as cr
+import feasibility_data.data.redcap.raw as dr
 
 SRC = Path(str(files("feasibility_data"))).joinpath("..").resolve()
 BLD = SRC.joinpath("..", "bld").resolve()
+RAW = SRC.joinpath("..", "raw").resolve()
 
 BLD_REDCAP = BLD / "redcap"
+RAW_REDCAP = RAW / "redcap"
```
signekb, line 16: As we get more and more of these, I think your suggestion of a class (#163) will become more and more relevant 😁

src/feasibility_data/build.py
signekb: Maybe add a TODO here?

signekb (changes requested):
This looks good to me. Just one suggestion about adding a TODO about handling the different centers :sunflower:

src/feasibility_data/build.py
signekb: Unless we split the build files into separate ones for each data source; `task_redcap.py`, `task_myfood24.py`, etc.

src/feasibility_data/data/redcap/raw.py
```diff
@@ -30,7 +30,7 @@ def download_data(center: cr.Center) -> str:
...
     data_path = raw_data_dir / f"{timestamp}.csv.gz"
-    raw_data_dir.parent.mkdir(parents=True, exist_ok=True)
+    raw_data_dir.mkdir(parents=True, exist_ok=True)
```
signekb, line 34: As I wrote in a comment in #173 yesterday, it seems that we don't need this (at least, I didn't for myfood24. I think that pytask creates the dir)

signekb (changes requested):
Very minor things:

src/feasibility_data/build.py
lwjohnst86: I haven't tried this but I thought Python allows something like `redcap.raw.download(...)`? And e.g. just `import feasibility_data.data.redcap as redcap`?

src/feasibility_data/data/redcap/raw.py
martonvago: Should we then remove this line and expect that this code will always run through pytask or keep it in case we ever want to run it outside of pytask? For debugging at least it can be nice to be able to run a function on its own, but it's no big deal either way.

src/feasibility_data/build.py
martonvago: Yeah, if `data/redcap/__init__.py` exposes the `raw` submodule. I'll think about what to call it though, because we have 3 `redcaps` (in `common`, `data`, `metadata`).

src/feasibility_data/data/redcap/raw.py
signekb: Hmm, yeah, I don't mind either actually. So maybe keep and I'll add in #173 as well so it's aligned?

martonvago: Okay! 🤝

src/feasibility_data/build.py
martonvago: In any case, we can rewrite the imports in another PR because it would touch the init files as well.

signekb: Agreed. I'll resolve this and make an issue about it. #178

src/feasibility_data/build.py
signekb: Will resolve this; see #179 instead.

signekb (dismissed):
✅ ☀️

signekb (dismissed):
I'll approve again 😁

src/feasibility_data/build.py
```diff
@@ -21,3 +24,17 @@ def task_download_field_metadata(
...
+        Path,
+        DirectoryNode(root_dir=RAW_REDCAP, pattern="*.csv.gz"),
+        Product,
+    ],
+) -> None:
+    """Download the latest data from all centers to `RAW_REDCAP/<timestamp>.csv.gz`."""
+    # TODO: Handle all centers
+    for center in [cr.Center.Copenhagen]:
+        data = dr.download_data(center)
+        dr.write_data(raw_data_dir, data)
```
lwjohnst86: This should be aligned to be "data first arg, output second arg", but that can be another PR.

src/feasibility_data/build.py
```diff
@@ -21,3 +24,17 @@ def task_download_field_metadata(
     """Download field metadata to `BLD_REDCAP`."""
     metadata = cr.get_json("metadata")
     cj.write_json(field_metadata_path, metadata)
+
+
+def task_download_data(
```
lwjohnst86:
```suggestion
def task_download_raw_redcap_data(
```

Since there will be other download of data tasks (e.g. myfood24).

lwjohnst86 (changes requested):
Very minor suggested change.

lwjohnst86 (approved):

PR #172, opened 2026-07-27
chore: 🍱 preprocess REDCap choices metadata during build

### LINKED ISSUE #168: Preprocess field metadata by expanding checkbox fields

martonvago:
# Description

This PR adds a task for preprocessing field metadata by expanding checkbox fields.
The code for expanding checkbox fields was adapted from the original `redcap_dict_to_properties.py`.

Why we still need to expand checkbox fields:
Before extracting properties from the data, we split the data by form/resource into separate data files. We can do this because the field metadata from REDCap lists which fields/cols go to which resource/form. For this to work, we need the field/col names in the metadata to match the names in the data. By default, this is not true for checkbox fields. In the metadata, checkbox fields are given as one field with a series of options under `select_choices_or_calculations`. In the data, each option gets its own column (with a value of checked or not checked). Same discussion [here](https://github.com/onlimit-study/feasibility-data/issues/32#issuecomment-4090670455).
So by replacing each checkbox field in the metadata with one field for each checkbox option, we get a field--form mapping that we can use to split the data.

Closes #168

Needs a thorough review.

## Checklist

- [x] Ran `just run-all`


src/feasibility_data/metadata/redcap/core.py
martonvago, line 1: See PR description for explanation

lwjohnst86 (approved):

PR #173, opened 2026-07-27
chore: 🍱 download raw myfood24 data during build

### LINKED ISSUE #134: Pull data from MyFood24 into raw
Confirm with @K-Beicher about API key, etc.

signekb:
# Description

Closes #134 

Needs a thorough review.

## Checklist

- [ ] Ran `just run-all`


src/feasibility_data/data/myfood24/raw.py
```diff
@@ -0,0 +1,42 @@
...
+    response = requests.post(api.url, headers=headers, stream=True)
+    response.raise_for_status()
+
+    return response
+
+
+def write(raw_data_dir: Path, response: requests.Response) -> None:
+    """Write the myfood24 data extract zip to the `raw/myfood24/` directory."""
+    timestamp = datetime.now(tz=UTC).strftime("%Y-%m-%dT%H-%M-%S")
+    data_path = raw_data_dir / f"{timestamp}.zip"
```
signekb: Are we fine with zip or should I unzip so we have compressed (but separate) csvs in raw instead?

src/feasibility_data/data/myfood24/raw.py
```diff
@@ -0,0 +1,42 @@
...
+    }
+    response = requests.post(api.url, headers=headers, stream=True)
+    response.raise_for_status()
+
+    return response
+
+
+def write(raw_data_dir: Path, response: requests.Response) -> None:
+    """Write the myfood24 data extract zip to the `raw/myfood24/` directory."""
+    timestamp = datetime.now(tz=UTC).strftime("%Y-%m-%dT%H-%M-%S")
```
signekb:
```suggestion
    timestamp = datetime.now(tz=UTC).strftime("%Y-%m-%dT%H-%M-%SZ")
```

signekb: Since this is also used for naming the REDCap downloads, we could move this to the common module?

src/feasibility_data/build.py
signekb, line 1: We could also have separate build pipelines per data source if it gets too messy with all in one file at some point? E.g., `task_redcap.py`, `task_myfood24.py`?

src/feasibility_data/data/myfood24/raw.py
```diff
@@ -0,0 +1,42 @@
+import os
+from datetime import UTC, datetime
+from pathlib import Path
+
+import requests
+from dotenv import load_dotenv
+
+from feasibility_data.common.redcap import APIConfig
```
martonvago: If we want to use `APIConfig` outside of `common.redcap`, it should probably live in something like `common.api` or `common.core`. But you don't necessarily have to use it here (you just wrap the values and then immediately unwrap them). For the REDCap stuff it's more useful because there are multiple centers. Will there only be one myfood24 project?

src/feasibility_data/data/myfood24/raw.py
martonvago: Yeah 👍

src/feasibility_data/data/myfood24/raw.py
martonvago: What do the separate CSVs correspond to? Different resources?

src/feasibility_data/data/myfood24/raw.py
```diff
@@ -0,0 +1,42 @@
...
+    response.raise_for_status()
+
+    return response
+
+
+def write(raw_data_dir: Path, response: requests.Response) -> None:
+    """Write the myfood24 data extract zip to the `raw/myfood24/` directory."""
+    timestamp = datetime.now(tz=UTC).strftime("%Y-%m-%dT%H-%M-%S")
+    data_path = raw_data_dir / f"{timestamp}.zip"
+    raw_data_dir.parent.mkdir(parents=True, exist_ok=True)
```
martonvago:
```suggestion
    raw_data_dir.mkdir(parents=True, exist_ok=True)
```
Right? Because we need `raw_data_dir` created as well. I will also fix this in the other `raw.py`, I think it was left over from when we were creating the parent of a path.

src/feasibility_data/data/myfood24/raw.py
```diff
@@ -0,0 +1,42 @@
...
+    with open(data_path, "wb") as f:
+        total_size = 0
+        for chunk in response.iter_content(chunk_size=8192):
+            if chunk:
+                f.write(chunk)
+                total_size += len(chunk)
```
martonvago:
```suggestion
        for chunk in response.iter_content(chunk_size=8192):
            if chunk:
                f.write(chunk)
```
Or do we need this for something?

src/feasibility_data/data/myfood24/raw.py
```diff
@@ -0,0 +1,42 @@
...
+        # and was flagged by gitleaks as a secret.
+        env_key="MYFOOD_API_KEY",
+        url="https://myfood24.org/api/projects/23816/extract",
+    )
+    headers = {
+        "Authorization": f"Api-Key {os.environ.get(api.env_key)}",
+        "Content-Type": "application/json",
+        "Accept": "application/json",
+    }
+    response = requests.post(api.url, headers=headers, stream=True)
```
martonvago: I think this will be flagged later for not having a timeout.

martonvago (changes requested):
Very nice, just minor things 🌞

src/feasibility_data/data/myfood24/raw.py
signekb: Yeah, I don't need it, just thought it was a little nicer to use the same as in REDCap - and potentially for the other data sources as well. But also fine with removing it. 

You're right: If we want to use it across sources, it should probably live in another file.

src/feasibility_data/data/myfood24/raw.py
signekb: Oh, flagged by what?

signekb:
```suggestion
    response = requests.post(api.url, headers=headers, stream=True, timeout=60)
```

martonvago: `bandit` in `check-security`

src/feasibility_data/data/myfood24/raw.py
signekb: Ooops, thanks!

src/feasibility_data/data/myfood24/raw.py
signekb: Ooooo, I actually just tried without the line entirely, and that seems to work as well; I think that maybe pytask creates the dir?

src/feasibility_data/data/myfood24/raw.py
signekb: Yup: 

- One with a "breakdown" of each thing they eat and drank, including the time, what meal, and a lot of stuff
- One with a "summary" with daily totals of what they eat and drank during a day, e.g., "total meat", protein, fat, sugar, etc.

signekb: I'll change this tomorrow to unzip and only save the compressed csvs instead 👍

src/feasibility_data/data/myfood24/raw.py
martonvago: Oh you're right! I guess it's because the path is declared as a product? But I wasn't able to find this described anywhere in their docs. So this only matters if the code runs outside of Pytask.

signekb: I think so? I also can't find it in their docs, but can just see that it works when running the pytask pipeline. 
Yes, but do we expect the code to ever be run outside pytask?

src/feasibility_data/data/myfood24/raw.py
signekb: Ah, oki 👌

src/feasibility_data/data/myfood24/raw.py
```diff
@@ -0,0 +1,58 @@
...
+def write(raw_data_dir: Path, response: requests.Response) -> None:
+    """Write the myfood24 data extract `.csv.gz` files to the raw directory."""
+    timestamp = datetime.now(tz=UTC).strftime("%Y-%m-%dT%H-%M-%SZ")
+
+    # Write zip to temporary file.
+    with tempfile.NamedTemporaryFile(suffix=".zip") as temp_file:
+        for chunk in response.iter_content(chunk_size=8192):
+            temp_file.write(chunk)
+        # Ensure data is written to disk.
+        temp_file.flush()
```
signekb, line 42: See e.g., https://www.geeksforgeeks.org/python/file-flush-method-in-python/

src/feasibility_data/data/myfood24/raw.py
```diff
@@ -0,0 +1,58 @@
...
+
+    return response
+
+
+def write(raw_data_dir: Path, response: requests.Response) -> None:
+    """Write the myfood24 data extract `.csv.gz` files to the raw directory."""
+    timestamp = datetime.now(tz=UTC).strftime("%Y-%m-%dT%H-%M-%SZ")
+
+    # Write zip to temporary file.
+    with tempfile.NamedTemporaryFile(suffix=".zip") as temp_file:
```
signekb, line 38: Not sure this is ideal, but thought we didn't want to keep the zip.

src/feasibility_data/data/myfood24/raw.py
```diff
@@ -0,0 +1,57 @@
...
+            temp_file.write(chunk)
+        # Ensure data is written to disk.
+        temp_file.flush()
+
+        # Write zip to gzip in `raw/`.
+        with zipfile.ZipFile(temp_file.name) as zip_file:
+            for file in zip_file.infolist():
+                output_path = (
+                    raw_data_dir
+                    / f"{Path(file.filename).stem}-{get_current_datetime()}.csv.gz"
```
martonvago: Maybe get the datetime once outside the loop, so all the files that come from one download are guaranteed to have the same timestamp.

src/feasibility_data/data/myfood24/raw.py
```diff
@@ -0,0 +1,57 @@
...
+def write(raw_data_dir: Path, response: requests.Response) -> None:
+    """Write the myfood24 data extract `.csv.gz` files to the raw directory."""
+    # Write zip to temporary file.
+    with tempfile.NamedTemporaryFile(suffix=".zip") as temp_file:
+        for chunk in response.iter_content(chunk_size=8192):
+            temp_file.write(chunk)
+        # Ensure data is written to disk.
+        temp_file.flush()
+
+        # Write zip to gzip in `raw/`.
```
martonvago:
```suggestion
        # Extract each CSV from the archive and recompress it as `.csv.gz`.
```
Maybe

martonvago (changes requested):
Nice, just minor comments ⭐

src/feasibility_data/build.py
signekb: See #179

src/feasibility_data/data/myfood24/raw.py
signekb: Oh, yeah, good point!

signekb: Changed 🤘

src/feasibility_data/data/myfood24/raw.py
```diff
@@ -0,0 +1,59 @@
...
+        # and was flagged by gitleaks as a secret.
+        env_key="MYFOOD_API_KEY",
+        url="https://myfood24.org/api/projects/23816/extract",
+    )
+    headers = {
+        "Authorization": f"Api-Key {os.environ.get(api.env_key)}",
+        "Content-Type": "application/json",
+        "Accept": "application/json",
+    }
+    response = requests.post(api.url, headers=headers, stream=True, timeout=60)
```
signekb: If we want to avoid using `APIConfig`:

````suggestion
    headers = {
        # Key couldn't be named "MYFOOD24" bc the entropy became too high and
        # and was flagged by gitleaks as a secret.
        "Authorization": f"Api-Key {os.environ.get("MYFOOD_API_KEY")}",
        "Content-Type": "application/json",
        "Accept": "application/json",
    }
    response = requests.post("https://myfood24.org/api/projects/23816/extract", headers=headers, stream=True, timeout=60)
```

martonvago: I don't mind either way, just the one comment about importing from `redcap`.

A halfway-house option would be putting the values into constants at the top.

signekb: I'll remove 🔥

src/feasibility_data/data/myfood24/raw.py
signekb: I've removed it 🔥

src/feasibility_data/data/myfood24/raw.py
signekb: I've added the line with a comment 👍

signekb:
All suggestions/discussions resolved (except my own comments).

src/feasibility_data/data/myfood24/raw.py
```diff
@@ -0,0 +1,61 @@
...
+        headers=headers,
+        stream=True,
+        timeout=60,
+    )
+    response.raise_for_status()
+
+    return response
+
+
+def write(raw_data_dir: Path, response: requests.Response) -> None:
```
lwjohnst86:
```suggestion
def write(response: requests.Response, output_dir: Path) -> None:
```

Output location should be the second argument, not first. Renamed to `output_dir` to keep is generic.

src/feasibility_data/data/myfood24/raw.py
```diff
@@ -0,0 +1,61 @@
...
+    """Write the myfood24 data extract `.csv.gz` files to the raw directory."""
+    # Create raw data dir. Strictly not necessary because pytask creates the dir, but
+    # useful if we ever need to run this outside the pytask pipeline.
+    raw_data_dir.mkdir(parents=True, exist_ok=True)
```
lwjohnst86:
```suggestion
    output_dir.mkdir(parents=True, exist_ok=True)
```

src/feasibility_data/data/myfood24/raw.py
```diff
@@ -0,0 +1,61 @@
...
+        # Ensure data is written to disk.
+        temp_file.flush()
+
+        # Extract each CSV from the archive and recompress it as `.csv.gz`.
+        with zipfile.ZipFile(temp_file.name) as zip_file:
+            current_datetime = get_current_datetime()
+
+            for file in zip_file.infolist():
+                output_path = (
+                    raw_data_dir
```
lwjohnst86:
```suggestion
                    output_dir
```

src/feasibility_data/build.py
```diff
@@ -21,3 +25,15 @@ def task_download_field_metadata(
...
+def task_download_myfood24_data(
+    myfood24_raw_data_dir: Annotated[
+        Path,
+        DirectoryNode(root_dir=RAW_MYFOOD24, pattern="*.zip"),
+        Product,
+    ],
+) -> None:
+    """Download the myfood24 data."""
+    data = dmr.download()
+    dmr.write(myfood24_raw_data_dir, data)
```
lwjohnst86:
```suggestion
    dmr.write(data, myfood24_raw_data_dir)
```

lwjohnst86 (changes requested):
nice! Just some small changes ☺️

martonvago (dismissed):

src/feasibility_data/data/myfood24/raw.py
```diff
@@ -0,0 +1,61 @@
...
+import tempfile
+import zipfile
+from pathlib import Path
+
+import requests
+from dotenv import load_dotenv
+
+from feasibility_data.common.datetime import get_current_datetime
+
+load_dotenv()
```
lwjohnst86:
```suggestion
```

src/feasibility_data/data/myfood24/raw.py
```diff
@@ -0,0 +1,61 @@
+import gzip
+import os
+import shutil
+import tempfile
+import zipfile
+from pathlib import Path
+
+import requests
+from dotenv import load_dotenv
```
lwjohnst86:
```suggestion
```

This should be in `build.py`, not here, since these files should only contain unevaluated functions. Can you move this into `build.py`?

lwjohnst86 (changes requested):

src/feasibility_data/data/myfood24/raw.py
signekb: On it 🫡

lwjohnst86 (approved):

PR #174, opened 2026-07-28
chore: 🔥 delete old build code

martonvago:
# Description

This PR deletes the old build (raw to staging) code. These were breaking the CI flow for various reasons and we won't need them / they will be refactored anyway.

Closes #

Needs a medium-depth review.

## Checklist

- [x] Ran `just run-all`


src/feasibility_data/metadata/core.py
martonvago, line 1: Kept the package properties, but removed the resource properties for now

signekb (approved):
Nice! 🔥

PR #176, opened 2026-07-29
chore: 🍱 add DK-EN mapping for choice values from REDCap

K-Beicher:
# Description

Just storing the translations of the selectable values in redcap until we need them for the data analysis extracts. 

Closes #74 

Needs a quick review.

## Checklist

- [ ] Ran `just run-all`


K-Beicher:
Got an error on just run-all, need to investigate that before I turn this into a PR.

lwjohnst86 (changes requested):
This should go under `src/feasibility_data/metadata/redcap/`. And also, file names need to always be lowercase

lwjohnst86 (changes requested):
The `redcap` in the name is redundant since it's in the redcap folder, so you can remove that from the name. And you could rename the last bit to `dk-to-en` since `en` is the abbreviation for English.

lwjohnst86 (changes requested):
You moved it into `metadata/redcap/`, but I said `src/feasibility_data/metadata/redcap/` (in the `src/` folder).

lwjohnst86 (approved):

PR #177, opened 2026-07-29
chore 🏗️: add common `get_current_datetime()` function

signekb:
# Description

So we can reuse across data sources when writing to `raw/`.

As suggested in a comment in #173.

### DIFF
--- a/src/feasibility_data/data/redcap/raw.py
+++ b/src/feasibility_data/data/redcap/raw.py
@@ -1,10 +1,10 @@
-from datetime import datetime
 from io import StringIO
 from pathlib import Path
 
 import polars as pl
 
 import feasibility_data.common.redcap as cr
+from feasibility_data.common.datetime import get_current_datetime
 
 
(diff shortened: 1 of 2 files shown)


src/feasibility_data/data/redcap/raw.py
```diff
@@ -1,10 +1,10 @@
-from datetime import datetime
 from io import StringIO
 from pathlib import Path
 
 import polars as pl
 
 import feasibility_data.common.redcap as cr
+from feasibility_data.common.datetime import get_current_datetime
```
martonvago, line 7: We'll have to decide if we want some system for these imports. As discussed [here](https://github.com/onlimit-study/feasibility-data/pull/171#discussion_r3672642621)

martonvago (approved):

src/feasibility_data/data/redcap/raw.py
signekb: Agreed. I've made an issue for it #178, so I'll resolve this.

PR #181, opened 2026-07-31
docs: 📝 describe REDCap metadata files

martonvago:
# Description

This PR describes the metadata files downloaded from REDCap.

Needs a quick review.

## Checklist

- [x] Ran `just run-all`


CONTRIBUTING.md
```diff
@@ -194,6 +194,25 @@ There are specific things to note about the REDCap data:
...
 `participant_id` and a `visit_id` field.
 
+#### REDCap metadata files
+
+Before we can extract properties from the data downloaded from REDCap, we have
+to tidy the data and split it into separate files by resource. To do this, we
+use the following metadata files downloaded from REDCap:
+
+- `field-metadata.json` (REDCap content `metadata`): The list of all fields
+  across all forms in the study. We use this to find which fields belong to
```
lwjohnst86: Can you make a glossary section for what "form" means? Or expand on here with an explanation? It still isn't clear. Like, is this a survey form that people enter data into? Who's the one entering the data? Are they all survey forms? Are all fields in the forms?

CONTRIBUTING.md
```diff
@@ -194,6 +194,25 @@ There are specific things to note about the REDCap data:
...
+  which form.
+- `event_metadata.json` (REDCap content `formEventMapping`): The list of all
+  form-event pairs. We use this to determine which forms are filled in at which
+  events (i.e., data collection points).
```
lwjohnst86: Can you add a glossary entry for event? Or expand on the explanation here with an example? What does that mean data collection point? Is that when the participant is present? Or whenever the data collector opens the survey form and fills in the items in that survey, regardless of whether a participant is there or not?

CONTRIBUTING.md
```diff
@@ -194,6 +194,25 @@ There are specific things to note about the REDCap data:
...
+- `repeating_forms_metadata.json` (REDCap content `repeatingFormsEvents`): The
+  list of all form-event pairs that includes only forms that can repeat, i.e.,
+  that can be submitted multiple times for the same participant at the same data
```
lwjohnst86: This explanation helps, thanks! Could you include an example of a survey form (or whatever form means) that allows multiple submissions?

lwjohnst86 (changes requested):
This already helps. I have some questions though 😛

CONTRIBUTING.md
```diff
@@ -194,6 +194,25 @@ There are specific things to note about the REDCap data:
 When processing the data, each resource should (almost always) contain a
 `participant_id` and a `visit_id` field.
 
+#### REDCap metadata files
+
+Before we can extract properties from the data downloaded from REDCap, we have
+to tidy the data and split it into separate files by resource. To do this, we
+use the following metadata files downloaded from REDCap:
+
+- `field-metadata.json` (REDCap content `metadata`): The list of all fields
```
signekb: I'm not sure what "REDCap content `metadata`" means. Can you elaborate? 🌷

signekb (commented):
Oh, super nice to have this written down 👌 I would also like some examples to help ground these concepts a bit.

CONTRIBUTING.md
martonvago: I hope that's clearer. It's the `content` value you specify in the request to the REDCap API.

CONTRIBUTING.md
```diff
@@ -252,3 +273,22 @@ and more complex than the non-build functions.
 - Each "public" function should be at the top of the module file, with "private"
   (prefixed with `_`) functions below them.
 - Classes, either public or private, go at the top of the file.
+
+## Glossary
```
martonvago, line 358: Feel free to put it somewhere else if you want.

The below are all general REDCap terms, they are not specific to the study (other than some examples).

CONTRIBUTING.md
```diff
@@ -252,3 +273,22 @@ and more complex than the non-build functions.
...
+  record information about a participant, such as demographics, laboratory
+  measurements, or questionnaire responses. Forms may be completed either by
+  participants (as surveys) or by members of the study team. Every field belongs
+  to exactly one form.
+- Event: In REDCap, an event is a scheduled data collection point in a
+  longitudinal study, such as Prescreening, Visit 1, or Phase 1. It represents a
+  planned stage of the study when data is collected from a participant (rather
+  than the date and time when data is entered into REDCap). One or more forms
+  can be assigned to each event to collect different kinds of information about
+  the participant.
```
martonvago, line 370: I hope this is clear, but lmk if not. 
Events are just the abstract entities REDCap gives you to build a longitudinal study. They don't say anything about the mechanics or timing of data entry. I can create a 3-year study with one event for each year and then log lots of data for Year 1 throughout the year or even after the year ends -- it will all be Year 1 data.

lwjohnst86: Wow, I'm glad you wrote this down, because I did NOT understand event in this way. I think, to match our domain, we should rename this to something more meaningful, because I will probably keep thinking of "event" as something else. But I'll think about a better name, plus I need to better understand how "event" is being used. E.g. from what I understand, "visit 1" would happen within "phase 1", so those should be two separate entities, not grouped together. But this all can be for another PR.

lwjohnst86 (dismissed):
Just a comment, but otherwise approved. I'll let @signekb review it before merging in.

CONTRIBUTING.md
martonvago: I'm very happy to call it something else in our own representation of the data and metadata! I'd only keep the REDCap terms while we're still in a REDCap context (so while we use these REDCap metadata files to transform the data).

Good point about the visit-phase confusion. However, if I understand it correctly, both phases and visits are REDCap events despite the fact that visits are contained within phases in the real-world study. Kris [writes](https://github.com/onlimit-study/dmp/issues/220#issuecomment-5117346937): "The challenge with the feasibility data is that the dieticians have decided to group their meetings based on the phases, whereas the study site visits are individual events." This may be challenging in other ways, but conceptually it's fine: some data is attached to a more general time window (a phase) and some data is attached to a more specific time window (a visit). Kris' suggestion to organise things by week should solve this.

martonvago: Raised #188 to address changing over to weeks

CONTRIBUTING.md
signekb: Ah, ok, thanks. Yes, it's clearer now.

CONTRIBUTING.md
```diff
@@ -252,3 +273,22 @@ and more complex than the non-build functions.
...
   (prefixed with `_`) functions below them.
 - Classes, either public or private, go at the top of the file.
+
+## Glossary
+
+- Form: In REDCap, a form or instrument is a collection of related fields that
+  record information about a participant, such as demographics, laboratory
+  measurements, or questionnaire responses. Forms may be completed either by
+  participants (as surveys) or by members of the study team. Every field belongs
+  to exactly one form.
```
signekb: Not sure if this is the right place, but as far as I understand, the fields must have unique names, e.g., there can't be multiple fields named  `timer_started` in different forms. It would then have to be  `timer_started_v10`,  `timer_started_v8` etc.

CONTRIBUTING.md
```diff
@@ -252,3 +273,22 @@ and more complex than the non-build functions.
...
+  longitudinal study, such as Prescreening, Visit 1, or Phase 1. It represents a
+  planned stage of the study when data is collected from a participant (rather
+  than the date and time when data is entered into REDCap). One or more forms
+  can be assigned to each event to collect different kinds of information about
+  the participant.
+- Repeating form: In REDCap, a repeating form is a form that can be completed
+  multiple times for the same participant within the same event. Each submission
+  represents a separate instance of the form for that participant and event. For
+  example, the Phase 1 Dietary Deviations form is a repeating form because a
+  participant can report multiple deviations during Phase 1 of the study.
```
signekb, line 375: How can we distinguish the different submissions? Do they have an ID field that describes what week/phase/event that they fall under?

CONTRIBUTING.md
signekb: Would it make sense to make a separate field item in this list?

signekb (commented):
Very nice to get this written down 🙌 
Three questions:

CONTRIBUTING.md
martonvago: Rows from repeating forms come with a `redcap_repeat_instance` field from REDCap, which we will rename to `submission_id`.

CONTRIBUTING.md
martonvago: Yes, fields must have unique names.

My feeling is that this is already detailed enough, but I can tack on the unique name thing here:

```suggestion
  participants (as surveys) or by members of the study team. Every field belongs
  to exactly one form and field names are unique across all forms.
```

CONTRIBUTING.md
signekb: Ah, oki. Thanks!

CONTRIBUTING.md
signekb: Super - I like that 👌

CONTRIBUTING.md
```diff
@@ -194,6 +194,27 @@ There are specific things to note about the REDCap data:
...
+#### REDCap metadata files
+
+Before we can extract properties from the data downloaded from REDCap, we have
+to tidy the data and split it into separate files by resource. To do this, we
+use the following metadata files downloaded from REDCap:
+
+- `field-metadata.json` (REDCap API `content` value: `metadata`): The list of
+  all fields across all forms in the study. We use this to find which fields
+  belong to which form.
+- `event_metadata.json` (REDCap API `content` value: `formEventMapping`): The
```
lwjohnst86, line 287:
```suggestion
- `event-metadata.json` (REDCap API `content` value: `formEventMapping`): The
```

CONTRIBUTING.md
```diff
@@ -194,6 +194,27 @@ There are specific things to note about the REDCap data:
...
+  list of all form-event pairs. We use this to determine which forms are filled
+  in at which events.
+- `repeating_forms_metadata.json` (REDCap API `content` value:
```
lwjohnst86, line 290:
```suggestion
- `repeating-forms-metadata.json` (REDCap API `content` value:
```

lwjohnst86 (changes requested):
I think we'll have to rename the event information at some point, but I need to think about that a bit longer. So I won't hold up the PR. Just a few minor edits.

CONTRIBUTING.md
```diff
@@ -194,6 +194,27 @@ There are specific things to note about the REDCap data:
 When processing the data, each resource should (almost always) contain a
 `participant_id` and a `visit_id` field.
 
+#### REDCap metadata files
+
+Before we can extract properties from the data downloaded from REDCap, we have
+to tidy the data and split it into separate files by resource. To do this, we
+use the following metadata files downloaded from REDCap:
+
+- `field-metadata.json` (REDCap API `content` value: `metadata`): The list of
```
lwjohnst86:
```suggestion
- `field_metadata.json` (REDCap API `content` value: `metadata`): The list of
```

martonvago: Oops

lwjohnst86 (approved):

Issue #8, opened 2025-12-21
Test how to collaborate on GenomeDK


lwjohnst86:

signekb:
We've agreed that @martonvago will look over seedcase-project/team#385 and test out how that works on GenomeDK with the `example-seedbeetle` repo.

martonvago:
I had a go at the above, and it seems to work just fine!

What I did: 
I cloned `example-seed-beetle` from GitHub into a random folder in my own user space. `example-seed-beetle` has some data files in a Git LFS store in a shared folder on GenomeDK, so I made sure to have Git LFS installed / the conda env active before cloning, so that the file pointers would be resolved. I could access the Git LFS store with no problems from my own clone -- including both reading and writing!

So basically all we need shared between people working on a project is the LFS store, we don't even necessarily need a shared clone of the repo. This would also avoid the `.venv` problem @signekb discovered (that the `.venv` points to the account of whoever first set up the project, and other people get permission errors when trying to run the project without recreating the `.venv` ).

One thing that is worth mentioning again is that each clone that fetches data files from the shared Git LFS store will also store local copies of those files. By default, Git LFS keeps downloaded objects in a local cache, and the files in the currently checked-out commit are also present in the working tree. As a result, each user can end up storing up to two copies of the data they have checked out. There are strategies for mitigating this (share the cache, fetch only the data you need), but some duplication seems inevitable.

martonvago:
Oh and we need to decide how to handle the `.env` file with the API keys. We could put it in a shared folder on GenomeDK and load it from there in each clone? When developing outside GenomeDK, we could still have a local `.env` for testing. We would first load the shared `.env`(if it exists) and then load the local one (if it exists).

lwjohnst86:
Those sound like good plans! Nice that each person can develop in their own space.

lwjohnst86:
We could include a `just clean-cache` recipe for the local cache thing

signekb:
That sounds neat. So, no repo in the shared folder, only the git-lfs folder and an `.env` file? 
And then load that file instead with `load_dotenv("/path/to/shared/project/")`? 

Seems like a part of the `.env` [best practices](https://env.dev/guides/env-vars-best-practices) is to include an `.env.example` either without values (just keys) or with placeholder values to serve as documentation, so we could consider that. 

From the best practices guide linked above:

[...]

PR #182, opened 2026-07-31
chore: 🔧 load `.env` from shared file

### LINKED ISSUE #180: Make sure `.env` file works locally and in individual clones on GenomeDK
> Oh and we need to decide how to handle the `.env` file with the API keys. We could put it in a shared folder on GenomeDK and load it from there in each clone? When developing outside GenomeDK, we could still have a local `.env` for testing. We would first load the shared `.env`(if it exists) and then load the local one (if it exists). 

 _Originally posted by @martonvago in [#8](https://github.com/onlimit-study/feasibility-data/issues/8#issuecomment-5088937048)_

martonvago:
# Description

This PR loads the env vars from a shared .env file.

Closes #180 

Needs a thorough review.

## Checklist

- [x] Ran `just run-all`


src/feasibility_data/common/dotenv.py
```diff
@@ -0,0 +1,11 @@
+from dotenv import load_dotenv
+
+
+def load_env_vars() -> None:
+    """Load the environment variables.
+
+    First, env vars are loaded from the shared project folder on GenomeDK.
+    Then, they are overwritten by values from the local .env file (if this exists).
```
martonvago, line 8: This allows us to use test APIs locally or test new APIs on GenomeDK

src/feasibility_data/common/dotenv.py
```diff
@@ -0,0 +1,11 @@
...
+    """
+    load_dotenv("/faststorage/project/sdca-onlimit-study/.env")
```
martonvago: Should I name it something else, e.g. `.feasibility-env`? Or put it in a folder?

.env.example
martonvago, line 1: As per https://github.com/onlimit-study/feasibility-data/issues/8#issuecomment-5130731366 ☺️

.env.example
```diff
@@ -0,0 +1,4 @@
+REDC_AAR_API_KEY=
+REDC_CPH_API_KEY=
+REDC_ODN_API_KEY=
```
signekb: Could this be a good time to rename these? Or should this be a separate PR? 
```suggestion
REDCAP_AAR_API_KEY=
REDCAP_CPH_API_KEY=
REDCAP_ODN_API_KEY=
```

.gitignore
```diff
@@ -20,6 +20,7 @@ venv/
 ENV/
 env.bak/
 venv.bak/
+!.env.example
```
signekb, line 23: Do we need to ignore this when it only contains the key names?

src/feasibility_data/common/dotenv.py
signekb: Oooh, maybe `.feasibility-env` is a good idea. Then we're prepared for when we need an `.env` for the main study. 
We could also add a folder where the main study env file can be placed as well, just so they're collected and not floating around in the  `sdca-onlimit-study` folder? E.g., `env/.feasibility-env` and then eventually `env/.main-env`?

```suggestion
    load_dotenv("/faststorage/project/sdca-onlimit-study/env/.feasibility-env")
```

signekb (changes requested):
Very nice 🚀 A couple of suggestions:

.gitignore
lwjohnst86: ? This line is saying *not* to ignore it

lwjohnst86 (commented):
Nice, agree with @signekb suggestions, otherwise approved!

.gitignore
martonvago: Yes, this is adding an exception to the blanket rule that all env files are ignored

signekb (approved):
✅ ✅

lwjohnst86 (approved):

PR #192, opened 2026-08-05
chore: 🍱 split REDCap data into one Parquet file per form

### LINKED ISSUE #169: Split raw data into into one Parquet file per form

martonvago:
# Description

This PR splits REDCap data into one Parquet file per form. I am saying "form" intentionally because moving to resources will only come in the next step. Some forms correspond one-to-one to resources, but some are first transformed or joined (e.g. vas). For the first kind, we can just move the parquet files generated here straight over to staging. For the second kind, we will do the necessary transformations when staging.

Of interest is the handling of metadata fields/columns that are not present in some data batches. This happens when new columns are added, which are missing in old batches. To make all staging data have the same shape, I add the missing columns in this step (filled with null). This should work out nicely because, when we join staging batches, newer batches will trump older batches. So values in the new columns in newer batches will trump the null placeholders.

For now, this only deals with `participant_id`, `event_id`, and `submission_id`. `study_week_id` and potentially other ids will fit roughly where `event_id` goes. There's already plenty in this PR to discuss, so I think it's simpler to let the `study_week_id` idea mature and refactor later. Structurally, (`participant_id`, `event_id`, `submission_id`) works as a PK, so we can continue building the pipeline in the meantime.

~I tested this on GenomeDK and it ran in 50ish seconds, so that's rather slow. Room for optimisation...~ down to 5 with some magic and sorcery 🧙 

Closes #169 

Needs a thorough review.

## Checklist

- [x] Ran `just run-all`


src/feasibility_data/build.py
```diff
@@ -77,6 +75,44 @@ def task_preprocess_field_metadata(
     common.json.write(field_metadata_preprocessed_path, field_metadata_preprocessed)
 
 
+def task_split_forms(
```
martonvago, line 87: I moved some complexity into this function as we wanted these tasks to do orchestration rather than just call another function that does the orchestration.

I do all reading and writing here to match previous tasks.

src/feasibility_data/build.py
```diff
@@ -77,6 +75,44 @@ def task_preprocess_field_metadata(
...
+    """
+    form_to_fields = data.redcap.core.get_form_field_mapping(
+        common.json.read(field_metadata_path)
+    )
+    form_to_events = data.redcap.core.get_form_event_mapping(
+        common.json.read(event_metadata_path)
+    )
+    repeating_form_names = data.redcap.core.get_repeating_forms(
+        common.json.read(repeating_forms_path)
+    )
```
martonvago: This is the metadata we will use to split the data.
I could unite them into a single structure if people like that better, e.g.:
```
[
  "bedq": {
    "fields": [...],
    "events": [...],
    "repeats": False,
  },
  ...
]
```
Could even be the output of the previous preprocessing step.

src/feasibility_data/data/redcap/core.py
```diff
@@ -0,0 +1,128 @@
...
+from operator import itemgetter
+from pathlib import Path
+
+import polars as pl
+import seedcase_soil as so
+
+
+@dataclass
+class Form:
+    """Class to hold the name and data of a form."""
```
martonvago: I decided to do this instead of writing the form name into each df as a separate column only to drop that column later, as it felt a bit cleaner.

src/feasibility_data/data/redcap/core.py
```diff
@@ -0,0 +1,128 @@
...
+    "record_id_s",
+    "redcap_event_name",
+    "redcap_repeat_instrument",
+    "redcap_repeat_instance",
+]
+
+
+def read_raw(raw_data_path: Path, form_to_fields: dict[str, list[str]]) -> pl.LazyFrame:
+    """Read the raw data into a LazyFrame with missing columns added."""
+    raw_lf = pl.scan_csv(raw_data_path, infer_schema=False)
```
martonvago, line 36: I'm reading into a lazy frame because the huge number of columns made the transformations very slow on data frames, even with only 2 raw batches. Lazy frames allow Polars to optimise operations a lot more.

src/feasibility_data/data/redcap/core.py
```diff
@@ -0,0 +1,128 @@
...
+    return _with_missing_columns(raw_lf, form_to_fields)
```
martonvago: I decided to add missing columns here, right at read-time. This means that we don't have to worry about this later on in the flow. Can be a separate step of course.

I also thought about extra columns (if a column is dropped later on in the study), but I don't think that's a problem. Any columns not in the latest metadata will not make it into staging, which feels like what we want.

Let me know if anyone has wise thoughts about these scenarios.

src/feasibility_data/data/redcap/core.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def get_form_field_mapping(
```
martonvago: These mapping functions use a for loop. I think this is the simplest and cleanest way of expressing the logic. But lmk if I should rewrite it.

src/feasibility_data/data/redcap/core.py
```diff
@@ -0,0 +1,128 @@
...
+    repeating_form_names: set[str],
+) -> Form:
+    form_name, field_names = form_entry
+    events = form_to_events.get(form_name, [])
+    is_repeating = form_name in repeating_form_names
+    content_fields = so.keep(field_names, lambda field: field not in REDCAP_ID_COLS)
+
+    columns = [
+        pl.col("record_id_s").alias("participant_id"),
+        pl.col("redcap_event_name").alias("event_id"),
```
martonvago, line 79: This is where the other ids would be set up. (See PR description.)

src/feasibility_data/data/redcap/core.py
```diff
@@ -0,0 +1,128 @@
...
+def _create_df_for_form(
+    form_entry: tuple[str, list[str]],
+    raw_lf: pl.LazyFrame,
+    form_to_events: dict[str, list[str]],
+    repeating_form_names: set[str],
+) -> Form:
+    form_name, field_names = form_entry
+    events = form_to_events.get(form_name, [])
+    is_repeating = form_name in repeating_form_names
+    content_fields = so.keep(field_names, lambda field: field not in REDCAP_ID_COLS)
```
martonvago: @signekb maybeee removing the admin fields could be part of this? We will include only the `content_fields` for each form, so we could strip out admin fields as well.

src/feasibility_data/data/redcap/core.py
```diff
@@ -0,0 +1,128 @@
...
+    filters = [
+        # Keep only rows coming from events where the form was filled in
+        pl.col("redcap_event_name").is_in(events),
+        # Keep only non-empty rows
+        pl.any_horizontal(
+            so.fmap(content_fields, lambda field: pl.col(field).is_not_null())
+        ),
+    ]
+
+    return Form(name=form_name, data=raw_lf.filter(filters).select(columns).collect())
```
martonvago: This is where the data frame is materialised (i.e. the operations are executed).

src/feasibility_data/data/redcap/core.py
lwjohnst86: This should be added as a comment, since it helps communicate why LazyFrame was used. Can you add that as a comment?

src/feasibility_data/data/redcap/core.py
lwjohnst86: It's hard to follow what's going on here and why these are needed, so it's hard for me to know what or if the loop is doing/how to improve on it. Could you refactor these to take a type/class and output a type/class that represents what you actually want/need from the forms? My instinct tells me this could be simpler but I can't pinpoint how yet.

src/feasibility_data/data/redcap/core.py
```diff
@@ -0,0 +1,128 @@
...
+    form_to_events: dict[str, list[str]],
+    repeating_form_names: set[str],
+) -> list[Form]:
+    """Split the raw data into one dataframe per form."""
+    forms = so.fmap(
+        form_to_fields.items(),
+        lambda form_entry: _create_df_for_form(
+            form_entry, raw_lf, form_to_events, repeating_form_names
+        ),
+    )
```
lwjohnst86: I believe that Polars has a way to do this rather than use maps, which will probably save a lot of time. A quick look seems that maybe `partition_by()` might be it..? I know it's possible to do this in R, so I assume Polars can do it to.

src/feasibility_data/data/redcap/core.py
lwjohnst86: You *might* not have to collect at this stage, as writing to file will force it. At least that's how it could be done in R, and I know Polars has lots of similarities. But it might be different.

src/feasibility_data/build.py
```diff
@@ -77,6 +75,43 @@ def task_preprocess_field_metadata(
...
+        raw_data = data.redcap.core.read_raw(raw_data_path, form_to_fields)
+        forms = data.redcap.core.split_forms(
+            raw_data,
+            form_to_fields,
+            form_to_events,
+            repeating_form_names,
+        )
+
+        for form in forms:
+            data.redcap.core.write_form(form, forms_dir, raw_data_path)
```
lwjohnst86, line 109: This can be rewritten with map, which might allow us to use some parallel processing from the refactor, which will help speed things up.

lwjohnst86 (changes requested):
Nice start 🎉 Here's some initial comments and suggestions, haven't fully reviewed every bit.

src/feasibility_data/data/redcap/core.py
martonvago: I think `partition_by` is for grouping rows by the values in one or more columns, not grouping columns based on their names. I wasn't able to find a Polars-native way of doing what we want, but if you know what exactly you would use in R, maybe we could check whether that has an equivalent.

src/feasibility_data/data/redcap/core.py
martonvago: Would [this suggestion](https://github.com/onlimit-study/feasibility-data/pull/192/changes#r3728819599) help?

src/feasibility_data/data/redcap/core.py
martonvago: We have to collect before writing to check if the df is empty because we said we didn't want to create empty resources. Buuut we could actually allow empty resources in staging and let the properties extraction and batch joining mechanisms deal with them later. Or there are other alternatives to the `is_empty` check, such as writing without checking and deleting afterwards if the Parquet metadata says 0 rows.

In any case, we should definitely optimise by collecting all frames at once.

src/feasibility_data/build.py
martonvago: There is a [paralleliser plugin](https://pytask-parallel.readthedocs.io/en/stable/) for Pytask, which could fit nicely here to parallelise the iterations of the outer loop (i.e. processing the raw batches). This would be equivalent to creating a map and executing in parallel.

Or do you mean something more specific by "the refactor"?

src/feasibility_data/build.py
signekb: Oh, that might read a little nicer/cleaner here, if they were combined in the previous step (or just in a single function in `data.redcap.core`,  but I don't have a strong opinion on this

src/feasibility_data/data/redcap/core.py
lwjohnst86: It might help, try it out :relaxed:

src/feasibility_data/build.py
lwjohnst86: Hmm, I couldn't see from the pytask paralleliser how that could parallelize *within* a task, just to parallelise all tasks.

I meant to refactor the loop into a map, so that we could e.g. use https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.multiprocessing.Pool.map which might make things faster.

martonvago: The paralleliser would work by declaring each iteration as a task and then parallelising all tasks whenever possible. Under the hood, by default it would use "a more robust implementation of `Pool` and `ProcessPoolExecutor`" [x](https://pytask-parallel.readthedocs.io/en/stable/quickstart.html#loky). So it would be kinda equivalent to using `Pool` or `ProcessPoolExecutor` manually.

The advantages are that the backend is more robust (if we trust that statement) and that orchestrating this in pytask would give us better progress tracking / visibility over the per-batch tasks. It would also give us pytask's dependency tracking mechanism for the batch files (i.e. only rerun if inputs changed), but we might not take advantage of that if we always force run the pipeline.

Another thing to be aware of here is that Polars is multithreaded by default, so we would be piling parallelism on parallelism here. Not necessarily bad, but we would want to limit the number of total threads to roughly the number of cores to avoid "oversubscription" (i.e., having so many threads that the OS can't divvy out one to each core, and instead has to constantly switch between them).

src/feasibility_data/data/redcap/core.py
martonvago: Done in #197

src/feasibility_data/build.py
martonvago: In any case, this can be decided and implemented separately -- I made #199

src/feasibility_data/data/redcap/core.py
martonvago: I'll resolve because I don't think there is a Polars equivalent, but lmk if you find one!

src/feasibility_data/data/redcap/core.py
```diff
@@ -0,0 +1,101 @@
...
+    return lf.with_columns(so.fmap(missing_fields, pl.lit(None).alias))
+
+
+def _create_df_for_form(
+    form_metadata: metadata.redcap.core.FormMetadata,
+    raw_lf: pl.LazyFrame,
+) -> pl.LazyFrame:
+    content_fields = so.keep(
+        form_metadata.fields, lambda field: field not in REDCAP_ID_COLS
+    )
```
martonvago, line 75: @signekb maybeee removing the admin fields could be part of this? We will include only the `content_fields` for each form, so we could strip out admin fields as well.

src/feasibility_data/data/redcap/core.py
```diff
@@ -0,0 +1,101 @@
...
+@dataclass(frozen=True)
+class Form:
+    """Class to hold the metadata and data of a form."""
+
+    form_metadata: metadata.redcap.core.FormMetadata
+    data: pl.DataFrame
+
+
+REDCAP_ID_COLS = [
+    "record_id_s",
```
signekb, line 19: Is it named `_s`?

src/feasibility_data/data/redcap/core.py
```diff
@@ -0,0 +1,101 @@
...
+    lf: pl.LazyFrame, forms_metadata: list[metadata.redcap.core.FormMetadata]
+) -> pl.LazyFrame:
+    """Add any missing metadata fields as columns in the dataframe."""
+    metadata_fields = so.flat_fmap(forms_metadata, lambda form: form.fields)
+    data_fields = set(lf.collect_schema().names())
+    missing_fields = so.keep(metadata_fields, lambda field: field not in data_fields)
+    return lf.with_columns(so.fmap(missing_fields, pl.lit(None).alias))
+
+
+def _create_df_for_form(
```
signekb: Do we want to go `lf` here to match the output type? 
```suggestion
def _create_lf_for_form(
```

src/feasibility_data/data/redcap/core.py
```diff
@@ -0,0 +1,101 @@
...
+
+    if form_metadata.repeats:
+        # Submissions for the same participant at the same event are told apart by
+        # `redcap_repeat_instance`.
+        columns.insert(
+            2, pl.col("redcap_repeat_instance").cast(pl.String).alias("submission_id")
+        )
+
+    filters = [
+        # Keep only rows coming from events where the form was filled in
```
signekb:
```suggestion
        # Keep only rows coming from events where the form was filled in.
```

signekb (changes requested):
Only very minor suggestions ✏️

src/feasibility_data/data/redcap/core.py
martonvago: Yes

signekb: How strange 🤔

martonvago: I've always thought it was for "string", but who knows 🤔

src/feasibility_data/data/redcap/core.py
martonvago: Oops, yes

src/feasibility_data/data/redcap/core.py
signekb: Oh, yeah, that could be 🤷 Just don't like the inconsistency. But I'll leave it be.

src/feasibility_data/data/redcap/core.py
```diff
@@ -0,0 +1,101 @@
...
+import seedcase_soil as so
+
+from feasibility_data import metadata
+
+
+@dataclass(frozen=True)
+class Form:
+    """Class to hold the metadata and data of a form."""
+
+    form_metadata: metadata.redcap.core.FormMetadata
```
lwjohnst86, line 14:
```suggestion
    form_metadata: metadata.redcap.core.Form
```

src/feasibility_data/data/redcap/core.py
lwjohnst86, line 19:
```suggestion
    # TODO: Rename to `participant_id` when ready
    "record_id_s",
```

src/feasibility_data/data/redcap/core.py
```diff
@@ -0,0 +1,101 @@
...
+REDCAP_ID_COLS = [
+    "record_id_s",
+    "redcap_event_name",
+    "redcap_repeat_instrument",
+    "redcap_repeat_instance",
+]
+
+
+def read_raw(
+    raw_data_path: Path, forms_metadata: list[metadata.redcap.core.FormMetadata]
```
lwjohnst86, line 27:
```suggestion
    path: Path, forms_metadata: list[metadata.redcap.core.Form]
```

src/feasibility_data/data/redcap/core.py
```diff
@@ -0,0 +1,101 @@
...
+) -> pl.LazyFrame:
+    """Read the raw data into a LazyFrame with missing columns added.
+
+    A LazyFrame lets Polars build a query plan and optimise it before executing, rather
+    than materialising intermediate results at each step. With a separate query for each
+    form and thousands of columns in the raw data, this meaningfully reduces computation
+    time compared to eager evaluation.
+    """
+    raw_lf = pl.scan_csv(raw_data_path, infer_schema=False)
```
lwjohnst86, line 36:
```suggestion
    raw_lf = pl.scan_csv(path, infer_schema=False)
```

src/feasibility_data/build.py
```diff
@@ -85,6 +84,31 @@ def task_create_form_metadata(
...
+
+    Written to `FORMS/<timestamp>/<form_name>.parquet`.
+    """
+    forms_metadata = so.fmap(
+        common.json.read(form_metadata_path),
+        lambda item: metadata.redcap.core.FormMetadata(**item),
+    )
+    for raw_data_path in raw_data_paths:
+        raw_data = data.redcap.core.read_raw(raw_data_path, forms_metadata)
+        forms = data.redcap.core.split_forms(raw_data, forms_metadata)
```
lwjohnst86, line 106: What I meant with the previous comment was to use map, e.g.:

```
out = map(raw_data_paths, read_and_split)
flat_out = flatten(out) # flatten one level
map(flat_out, write_form)
```

src/feasibility_data/data/redcap/core.py
```diff
@@ -0,0 +1,101 @@
...
+
+def split_forms(
+    raw_lf: pl.LazyFrame,
+    forms_metadata: list[metadata.redcap.core.FormMetadata],
+) -> list[Form]:
+    """Split the raw data into one dataframe per form."""
+    lazy_dfs = so.pairwise_fmap(forms_metadata, [raw_lf], _create_lf_for_form)
+    dfs = pl.collect_all(lazy_dfs)
+    forms = so.pairwise_fmap(forms_metadata, dfs, Form)
+    return so.keep(forms, lambda form: not form.data.is_empty())
```
lwjohnst86, line 48: My intuition is buzzing here that we could do this more simply, but I can't put my finger on it. I think, if we append the form name to the start of each column name (as an abbrev e.g. `ae_` for adverse events), we map-split by abbreviation so each form gets created into its standalone data frame. Which I think means we could cut down on the other code below.

lwjohnst86 (changes requested):
Some initial comments and changes, that will change later code (which I haven't reviewed).

src/feasibility_data/data/redcap/core.py
martonvago: This is coming from REDCap, we do rename it when we split. Do you mean rename in REDCap?

src/feasibility_data/data/redcap/core.py
martonvago: I'm very open to doing it differently, but I'm not sure I understand what you are suggesting.

- To append the form name to each column name, we would need to know which field/column belongs to which form. So we would still need to construct a mapping like `forms_metadata`.
- If we had the form name as a prefix in each column name, splitting would work by (1) extracting the prefix from each, (2) grouping column names by it, and (3) then calling `raw_lf.select(columns)` for each group/form. This is pretty much what we do already, but with the added prep steps of 1 and 2.
- This only addresses finding which column belongs to which form. The other bits in `_create_lf_for_form` would still need to be done (keeping only contentful columns, adding our own columns, renaming columns, filtering out irrelevant rows for each form).

src/feasibility_data/build.py
martonvago: We can definitely refactor it like this but I'm not sure I see how that's better.

- If the goal is to parallelise the processing of raw files, then the parallelisable unit is all the processing done to one raw file. In your suggestion, reading+splitting is divorced from writing. So if we parallelise reading+splitting, we would need to collect all results from these parallel processes, hold them all in memory, and then do a serial write. That strikes me as worse from both a parallel-to-serial overhead and a memory perspective than parallelising the whole read-split-write flow.
- I think the second/inner loop shouldn't be a map because it writes to file, so it doesn't really map to anything.
- Writing it like this also says no to parallelising with Pytask. We can make that decision now but then we should resolve the [earlier discussion](https://github.com/onlimit-study/feasibility-data/pull/192/changes#r3766889114) about it.

lwjohnst86:
Me and @K-Beicher are working on a mapping between field names and resources (plus updating the field names in the process), which will probably mean that this PR and the one stacked below will require heavy updates. We'll keep this on hold for now. I'll mark as draft to indicate the status.

PR #195, opened 2026-08-10
chore: 🔧 don't check csv content for typos, only file names

signekb:
# Description

Based on the [example](https://github.com/crate-ci/typos#false-positives) in typos' README.

This way, we still check the file names for typos and we don't have to exclude `.csv`s individually 

Needs a quick review.


lwjohnst86 (approved):
Nice!

PR #197, opened 2026-08-13
chore: :bento: create form metadata in preprocess step

martonvago:
# Description

This PR refactors the metadata preprocessing step to take all metadata files we download from REDCap and join them into one metadata file that lists each REDCap form with only the info we need to then split the data by form (form name, fields, events, whether the form repeats).

Arose from https://github.com/onlimit-study/feasibility-data/pull/192/changes#r3728819599 .

Closes #

Needs a thorough review.

## Checklist

- [x] Ran `just run-all`

The created form metadata looks like:
```json
[
  {
    "name": "stamdata",
    "fields": [
      "record_id_s",
      "id_number_pre",
      "name_stamdata",
      "date_of_birth_stamdata",
      ...
    ],
    "events": [
      "stamdata_arm_1"
    ],
    "repeats": false
  },
  {
    "name": "adverse_events",
    "fields": [
      "ae_report_date",
      "ae_event",
      ...
    ],
    "events": [
      "stamdata_arm_1"
    ],
    "repeats": true
  },
  {
    "name": "bookinger",
    "fields": [
      "book_v1",
      ...
    ],
    "events": [
      "stamdata_arm_1"
    ],
    "repeats": false
  },
  {
    "name": "prscreening_telefoninterview_frste_kontakt",
    "fields": [
      "date_v0",
      "staff_initials_v0",
      ...
    ],
    "events": [
      "prscreening_arm_1"
    ],
    "repeats": false
  },
  {
    "name": "randomisering",
    "fields": [
      "randomisation_yn",
      "is_randomised",
      "rand_datetime",
      "rand_datetime77",
      "randomisation",
      "isnot_randomised"
    ],
    "events": [
      "randomisering_arm_1"
    ],
    "repeats": false
  },
  {
    "name": "fr_besgsdag_1_screening",
    "fields": [
      "bloodsamples_order_wfv1",
      "cgm_ready_wfv1",
      "cgm_zip_wfv1",
      "sens_ready_wfv1",
      "sens_sign_in_wfv1",
      "liva_ready_wfv1"
    ],
    "events": [
      "besg_1__screening_arm_1"
    ],
    "repeats": false
  },
  ...
]
```


src/feasibility_data/metadata/redcap/core.py
```diff
@@ -1,9 +1,48 @@
...
+    name: str
+    fields: list[str]
+    events: list[str]
+    repeats: bool
+
+
+def create_form_metadata(
+    field_df: pl.DataFrame,
+    event_df: pl.DataFrame,
+    repeating_forms_df: pl.DataFrame,
```
martonvago, line 22: Taking these as data frames to transform and join them in Polars, avoiding the rather iterative handling in plain Python. I wouldn't necessarily do this if we weren't already using Polars, but we are!

src/feasibility_data/metadata/redcap/core.py
```diff
@@ -1,9 +1,48 @@
...
+) -> list[dict[str, Any]]:
```
martonvago, line 23: I'm returning stuff in the structure of `FormMetadata`, but I'm not instantiating it because it will just be written to file as JSON in `build.py`. I could return `FormMetadata` for clarity though.

src/feasibility_data/metadata/redcap/core.py
```diff
@@ -1,9 +1,48 @@
...
-from typing import cast
+from dataclasses import dataclass
+from typing import Any, cast
 
+import polars as pl
 import seedcase_soil as so
 
 
+@dataclass(frozen=True)
+class FormMetadata:
```
lwjohnst86, line 10: Since this is already within `metadata` folder, you can simplify to `Form`


```suggestion
class Form:
```

I'm also a bit confused, I can't seem to find where this is used. Could you point that out?

src/feasibility_data/metadata/redcap/core.py
```diff
@@ -1,9 +1,48 @@
...
+    """Class to hold the metadata of a form."""
+
+    name: str
+    fields: list[str]
+    events: list[str]
+    repeats: bool
+
+
+def create_form_metadata(
```
lwjohnst86, line 19:
```suggestion
def create_form(
```

Same point here.

src/feasibility_data/metadata/redcap/core.py
lwjohnst86, line 23:
```suggestion
) -> Form:
```

src/feasibility_data/build.py
```diff
@@ -63,18 +64,25 @@ def task_download_raw_redcap_data(
...
-    """Preprocess field metadata."""
+    """Create the metadata file describing each form."""
+    event_metadata = pl.read_json(event_metadata_path)
+    repeating_forms_metadata = pl.read_json(repeating_forms_path)
     field_metadata = common.json.read(field_metadata_path)
-    field_metadata_preprocessed = metadata.redcap.core.expand_checkbox_fields(
-        field_metadata
+    field_metadata = metadata.redcap.core.expand_checkbox_fields(field_metadata)
+
+    form_metadata = metadata.redcap.core.create_form_metadata(
```
lwjohnst86, line 79:
```suggestion
    form_metadata = metadata.redcap.core.create_form(
```

src/feasibility_data/build.py
```diff
@@ -63,18 +64,25 @@ def task_download_raw_redcap_data(
...
-def task_preprocess_field_metadata(
-    field_metadata_preprocessed_path: Annotated[
-        Path, Product
-    ] = FIELD_METADATA_PREPROCESSED_PATH,
+def task_create_form_metadata(
+    form_metadata_path: Annotated[Path, Product] = FORM_METADATA_PATH,
     field_metadata_path: Path = FIELD_METADATA_PATH,
+    event_metadata_path: Path = EVENT_METADATA_PATH,
+    repeating_forms_path: Path = REPEATING_FORMS_METADATA_PATH,
 ) -> None:
```
lwjohnst86, line 72:
```suggestion
@pytask.mark.metadata
def task_create_form_metadata(
    form_metadata_path: Annotated[Path, Product] = FORM_METADATA_PATH,
field_metadata_path: Path = FIELD_METADATA_PATH,
    event_metadata_path: Path = EVENT_METADATA_PATH,
    repeating_forms_path: Path = REPEATING_FORMS_METADATA_PATH,
) -> None:
```

Add a marker to help with managing build steps.

lwjohnst86 (changes requested):
Nice ☺️ just some minor changes plus a question

src/feasibility_data/metadata/redcap/core.py
martonvago: I thought about calling it just `Form`, but because it's used in `data.redcap.core.Form` (#192), I thought it was a bit confusing to be nesting `Form`s, even with `metadata` in the module path. E.g. in the IDE:

Image

vs

Image

If you don't mind this, I'll rename though!

martonvago: > I'm also a bit confused, I can't seem to find where this is used. Could you point that out?

This is used when reading in the file created here. I included it in this PR (rather than #192) because it sets out the structure of the metadata in the file and because I wanted to ask the [question below.](https://github.com/onlimit-study/feasibility-data/pull/197/changes#r3774070511)

src/feasibility_data/metadata/redcap/core.py
martonvago: Does this mean you prefer returning `FormMetadata`? In that case would it make sense to add a small dependency that handles dataclass <-> JSON conversion for us (e.g., [cattrs](https://pypi.org/project/cattrs/))? 

I was already considering this in #192 to get a more robust read behaviour, but there's not much of a difference when writing.

pyproject.toml
```diff
@@ -48,6 +48,7 @@ strict = true
 [tool.pytask.ini_options]
 paths = ["src/feasibility_data"]
 task_files = ["task_*.py", "build.py"]
+markers = ["metadata"]
```
martonvago: Without this it was complaining that "UserWarning: Unknown pytask.mark.metadata - is this a    │
│     typo? You can register custom marks to avoid this warning. "

pyproject.toml
```diff
@@ -45,9 +45,14 @@ extend = ".config/ruff.toml"
 python_version = "3.12"
 strict = true
 
+[[tool.mypy.overrides]]
+module = "feasibility_data.build"
+disallow_untyped_decorators = false
```
martonvago: The `mark` decorator is untyped. I added this build-specific rule because presumably all tasks will have a marker.

